PR #8, opened 2021-02-15
Adding team.blue DK mirror

fbhdk:


jaboutboul:
@fbhdk your email seems to be part of a distribution group which only accepts list members. Can you please provide a valid email address or add a wildcard for *almalinux.org?

fbhdk:
> @fbhdk your email seems to be part of a distribution group which only accepts list members. Can you please provide a valid email address or add a wildcard for *almalinux.org?

Oops! Sorry for that. It has been fixed now so external senders can send to it.

PR #15, opened 2021-02-17
LU-2208: AlmaLinux mirrors support improvement

soksanichenko:
- Big refactoring
- Both scripts are merged into one
- We check last update date of a mirror:
-- An expired mirror has status `expired` in mirrors table
-- An expired mirror isn't available in mirrorlist


mirrors_update.py
```diff
@@ -0,0 +1,318 @@
...
+def get_config(path_to_config: AnyStr = 'config.yml') -> Dict:
+    """
+    Read, parse and return mirrorlist config
+    """
+
+    with open(path_to_config, mode='r') as config_file:
+        return yaml.safe_load(config_file)
+
+
+def mirror_available(
```
andrewlukoshko, line 41: repo.almalinux.org must be always "available"

mirrors_update.py
```diff
@@ -0,0 +1,318 @@
...
+                mirror_info['name']
+            )
+            addresses = mirror_info['address']
+            for protocol in ALL_MIRROR_PROTOCOLS:
+                if protocol in addresses:
+                    link = f'[{address_prefixes[protocol]}]' \
+                           f'({addresses[protocol]})'
+                else:
+                    link = ''
+                mirror_info[f'{protocol}_link'] = link
```
andrewlukoshko, line 273: if link ends with '/' symbol we should remove it due to vuepress behavior when processing rsync:// links
but it'll never hurts to remove '/' from http/https addresses so it can be done for every link

PR #18, opened 2021-02-18
add new mirror ftp.fau.de

rrzefox:
two comments:
1. <strike>your mirroring-howto does not mention a recommended mirrorinterval, so for now we settled on _our_ default of 6 hours.</strike> no idea how I managed to overlook that, it does say 3 hours, I've updated both our config and the PR.
2. I don't see any 'country' or at least 'continent' field in the mirror infos, but you'll probably need that sooner or later to pre-filter the list of mirrors offered to a client as the mirrorlist gets larger and larger? Or do you plan to do that only via geoip?

### DIFF
--- a/mirrors.d/ftp.fau.de.yml
+++ b/mirrors.d/ftp.fau.de.yml
@@ -1,0 +1,11 @@
+---
+name: ftp.fau.de
+address:
+  http: http://ftp.fau.de/almalinux/
+  https: https://ftp.fau.de/almalinux/
+  rsync: rsync://ftp.fau.de/almalinux/
+update_frequency: 3h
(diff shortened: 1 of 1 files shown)


andrewlukoshko:
Hello and thanks for the mirror!
Currently we have plan to use geoip to simplify mirror submission process.

jaboutboul:
@rrzefox your email is returning as "Recipient address rejected: User unknown." Can you please provide a valid email address?

rrzefox:
Sorry, I typoed, there is an 's' missing - it's rrze-ftp-admin**s**@. Do you want another PR to fix the typo?

jaboutboul:
yes please. ty 😉

PR #20, opened 2021-02-18
Create mirror.cov.ukservers.com.yml

IGSteven:
We have added this to our Mirror located in the United Kingdom.

We will be adding HTTPS & RSYNC at a later date but will do a separate request for those.


IGSteven:
I have made an error in this request and will resubmit.

PR #25, opened 2021-02-19
Add datakeepers.co.za mirror

rjobber:


andrewlukoshko:
Hello and thanks for mirror!
Something is wrong however as I don't see 8 and 8.3 symlinks and almalinux-release-latest-8.x86_64.rpm file like on [official mirror](https://repo.almalinux.org/almalinux/).
Looks like they're exist but forbidden.
Did you follow [mirror setup guidelines](https://wiki.almalinux.org/Mirrors.html)?

rjobber:
Hi
Guidelines were followed, however symlinks were forbidden in web config, which was overlooked.
This has been corrected, sorry about that!
Robin

PR #29, opened 2021-02-20
Add "allowed_outdate" config option for mirrors age checking

andrewlukoshko:


mirrors_update.py
```diff
@@ -174,7 +178,7 @@ def get_verified_mirrors(
                 versions=versions,
                 repos=repos,
             ):
-                set_repo_status(mirror_info)
+                set_repo_status(mirror_info,allowed_outdate)
```
soksanichenko, line 181: pep8

PR #30, opened 2021-02-20
altushost mirror included

AltusHost:
We have added the almalinux mirror which will be served from our NL location.

Kindly review and add it.


AltusHost:
It is still in pending state. Can you please review?

AltusHost:
@andrewlukoshko  Hi,  The mirror status is showing as expired here https://mirrors.almalinux.org/

Should I recheck anything at our mirror server, can you please let me know.

PR #51, opened 2021-03-04
Add new mirror

RetroDawg:
New mirror yum.tamu.edu

### DIFF
--- a/mirrors.d/yum.tamu.edu.yml
+++ b/mirrors.d/yum.tamu.edu.yml
@@ -1,0 +1,10 @@
+---
+name: yum.tamu.edu
+address:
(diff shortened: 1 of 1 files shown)


jaboutboul:
what up @RetroDawg? The email you provided is some sort of distribution list which does not allow external senders. Can you please either add a wildcard to allow from almalinux.org or add another valid email address? Thanks!

jaboutboul:
@RetroDawg Hi Jose, we did not receive any response regarding this issue.

jonathanspw:
@RetroDawg it looks like the mirror is completely offline, not even having DNS anymore.  Will it be making a return?

PR #53, opened 2021-03-05
LU-2219: AlmaLinux mirrors impovement

soksanichenko:


src/backend/api/handlers.py
```diff
@@ -0,0 +1,191 @@
...
+logger = get_logger(__name__)
+
+
+def _get_distance_between(
+        lat1: float,
+        lon1: float,
+        lat2: float,
+        lon2: float,
+) -> float:
+    return math.sqrt(
```
Monstrofil, line 35: Just as a little tip: sqrt is quite expensive operation and you don't actually need the distance itself to sort mirrors by distance. So you can easily drop that math.sqrt call and return squared value and sort by it.

src/backend/api/handlers.py
```diff
@@ -0,0 +1,191 @@
...
+                Mirror.is_expired == false(),
+                ).all()
+            all_mirrors = [mirror.to_dict() for mirror in all_mirrors]
+            return all_mirrors
+        match_dict = match.get_info_dict()
+        match_country = match_dict['country']['names']['en']
+        match_continent = match_dict['continent']['names']['en']
+        suitable_mirrors = session.query(Mirror).filter(
+            Mirror.continent == match_continent,
+            Mirror.country == match_country,
```
Monstrofil, line 55: I guess this is "and" query? That may explain why I get Indonesian mirror when I'm requesting mirrors from UA :)

Maybe use some kind of UNION query where we first look for country and then for continent? 
Like
SELECT * FROM ... WHERE country = 'ua'
UNION SELECT * FROM ... WHERE continent = 'europe'
LIMIT 10

src/backend/api/handlers.py
```diff
@@ -0,0 +1,191 @@
...
+                os.path.abspath(__file__),
+            ),
+            '../../../mirrors',
+            config['mirrors_dir'],
+        ),
+        versions=versions,
+        repos=repos,
+        allowed_outdate=config['allowed_outdate']
+    )
+    with session_scope() as session:
```
Monstrofil, line 109: Does session_scope apply atomic() inside? Because otherwise we may remove all mirrors and don't get them back.

src/backend/api/handlers.py
```diff
@@ -0,0 +1,191 @@
...
+                email=mirror['email'],
+                urls=urls_to_create,
+            )
+            session.add(mirror_to_create)
+        session.flush()
+
+    return 'Done'
+
+
+def get_all_mirrors():
```
Monstrofil, line 141: You have an import of same method btw.

src/backend/api/mirrors_update.py
```diff
@@ -0,0 +1,236 @@
...
+def set_geo_data(
+        mirror_info: Dict[AnyStr, Union[Dict, AnyStr]],
+) -> None:
+    """
+    Set geo data by IP of a mirror
+    :param mirror_info: Dict with info about a mirror
+    """
+
+    mirror_name = mirror_info['name']
+    ip = socket.gethostbyname(mirror_name)
```
Monstrofil, line 214: Something like missing payment for domain or missing A record may break the whole tool, need some error handling here.

src/backend/api/handlers.py
```diff
@@ -0,0 +1,191 @@
...
+            rest_mirrors = session.query(Mirror).filter(
+                Mirror.name.notin_(
+                    [mirror.name for mirror in suitable_mirrors]
+                ),
+                Mirror.is_expired == false(),
+            ).all()
+            rest_mirrors = [mirror.to_dict() for mirror in rest_mirrors]
+        distance_dict = {
+            _get_distance_between(
+                lat1=match.location[0],
```
Monstrofil, line 79: Btw you gave me an idea 

Django ORM is able to return calculated field, so can SQLAlchemy I guess. You can move this ordering and sorting to postgresql (or even check postgis, it has better indexes and methods for lat/lon mathmatics). That will be much faster with indexes than python code.

May be postponed to the future updates.

src/backend/api/handlers.py
```diff
@@ -0,0 +1,191 @@
...
+        lon1: float,
+        lat2: float,
+        lon2: float,
+) -> float:
+    return math.sqrt(
+        (lat1 - lat2)**2 + (lon1 - lon2)**2
+    )
+
+
+def _get_nearest_mirrors(ip_address):
```
Monstrofil, line 40: Adding docstring descibing this logic would be really cool

Monstrofil (changes requested):
Hi, just checked your pull request, please, check comments that I left.

src/backend/api/handlers.py
soksanichenko: Done. It's replaced by hybrid method in the model

src/backend/api/handlers.py
soksanichenko: Done. Replaced by union request.

src/backend/api/handlers.py
soksanichenko: Done. Now it's used hybrid method.

src/backend/api/mirrors_update.py
soksanichenko: It's added try...except

src/backend/api/handlers.py
soksanichenko: It's added.

src/backend/api/handlers.py
soksanichenko: Yes, I'm applying a transaction after all operations. And I disabled autoflush.

src/backend/main.py
```diff
@@ -0,0 +1,180 @@
...
+)
+def mirrors_table():
+    url_types = sorted(get_url_types())
+    data = {
+        'column_names': [
+            'Name',
+            'Sponsor',
+            'Status',
+            'Continent',
+            'Country',
```
andrewlukoshko, line 129: Please use Location instead of Country

src/backend/templates/isos.html
```diff
@@ -0,0 +1,56 @@
...
+            <h3>You can also choose from full list of mirrors:</h3>
+        {% endif %}
+{% endblock text %}
+
+{% block table %}
+{{ super() }}
+    <table id="dtBasic" class="table table-striped table-bordered table-responsive-sm" data-order='[[ 0, "asc" ]]'>
+        <thead>
+            <tr>
+                <th>Country</th>
```
andrewlukoshko, line 34: Please use Location instead of Country

andrewlukoshko (commented):
Location should be used instead of Country in mirrors and ISO tables

PR #56, opened 2021-03-06
Update mirrorlists

mertcenikut:


mertcenikut:
Update mirrorlists

andrewlukoshko:
Hello and thanks for mirror.
But it looks like it's not updating: https://mirror-denizli.cenuta.com/almalinux/timestamp.txt
Did you follow mirrors guidelines? https://wiki.almalinux.org/Mirrors.html

mertcenikut:
Hello,
We fixed cron problem. Could you check again please?

PR #60, opened 2021-03-09
Create mirror-id.almalinux.id.yml

malhuda2:


jaboutboul:
@malhuda2 your email is returning as "Recipient address rejected: User unknown." Can you please provide a valid email address?

jaboutboul:
@malhuda2 we have not received any response here.

PR #63, opened 2021-03-16
New Frankfurt Mirror proposal

SolSoCoG:
Hello,
this request adds a mirror in Germany, Frankfurt with 10Gbps Bandwidth and DDoS filters.

/dev/null/v is commited in supporting open source projects with space and by running live examples of them.
The recommended 3h update cron is set.
This mirror also fully supports IPv6, which unfortunately still seems to be a rarity.

Best regards,
Alex


titan-dn (commented):
approve

PR #65, opened 2021-03-19
add amscloud mirrors

udamuri:
Add mirror amscloud

Kind regards


andrewlukoshko:
Hello and thanks for the mirror.
But it looks like it's not automatically updating: https://mirror.amscloud.co.id/almalinux/timestamp.txt
Did you setup cron?

andrewlukoshko:
Looks like everything is OK now. Merging.

PR #67, opened 2021-03-25
Create almalinux.mirror.letscloud.io.yml

marciorubens:


andrewlukoshko:
Hello and thank you for mirror!
Looks like it's not fully synced so could you check it?

marciorubens:
I performed rsync again.
Please check if it was successful

PR #68, opened 2021-03-25
NL mirror added (mirror.mijn.host)

egogo-nl:


egogo-nl:
Hello,
Our mirror was down because of infrastructure change.
Now the mirror is working again, and will stay working.
Can you re-add our mirror?

PR #70, opened 2021-03-29
Create mirrors.layerbridge.com.yml

AGSQ11:
Add LayerBridge (AS3280) RO mirrors.


AGSQ11 (commented):
Fixed wrong rsync url format.

PR #77, opened 2021-03-31
Create ftp.arnes.si.yml

mmihelic13:
Description of the AlmaLinux mirror at ARNES

### DIFF
--- a/mirrors.d/ftp.arnes.si.yml
+++ b/mirrors.d/ftp.arnes.si.yml
@@ -1,0 +1,11 @@
+---
+name: ftp.arnes.si
+address:
+  http: http://ftp.arnes.si/mirrors/almalinux
+  https: https://ftp.arnes.si/mirrors/almalinux
+  rsync: ftp.arnes.si::almalinux
+update_frequency: 6h
+sponsor: ARNES
+sponsor_url: https://www.arnes.si
+email: ftp-admin@arnes.si
+...


andrewlukoshko:
Hello and thanks for mirror!
It looks like it's not updating, did you setup cron?

mmihelic13:
No cronjobs have been run yet. The initial sync was completed yesterday somewhere between 22:00-23:00. I enabled the cronjob this morning, it is set to `22 1,7,15,21`. I can run a manual rsync now if needed and then the cronjob will do its thing at `15:22` and every 6 hours from then on.

andrewlukoshko:
OK, we'll wait for 13:22 then. No need to update it manually.

mmihelic13:
Cron job ran as scheduled, but the `TIME` and `timestamp` files were not updated. I will run the cronjob manually to see if what the output is.

mmihelic13:
When using the cronjob proposal from the mirrors web page, we get an error
```
# /usr/bin/flock -n /var/run/almalinux_rsync.lock -c /usr/bin/rsync -avSH -f 'R .~tmp~' --delete-delay --delay-updates rsync://rsync.repo.almalinux.org/almalinux/ /REAL_PATH_HERE/mirrors/almalinux/
flock: -c requires exactly one command argument
```
The man page for `flock` on CentOS7 says
```
       -c, --command command
              Pass a single command, without arguments, to the shell with -c.
```
Passing the whole `rsync` command with arguments will not work. I will remove the `flock` call and use `rsync` directly.

Next run is at `19:22`.

andrewlukoshko:
Sorry for that.
Guide is now fixed: https://wiki.almalinux.org/Mirrors.html

PR #86, opened 2021-04-01
mirror.fidonet.io

joacimmelin:


andrewlukoshko:
Hello and thanks for mirror.
Looks like your mirror has restrictions for some countries as I was able to access is from one servers and wasn't from other.
Is it possible to disable such restriction if it really enabled?

joacimmelin:
I've blocked countries that are either frequent attackers to my servers, or run by dictators (ie: Russia, China, North Korea, etc).   What country where you not able to access it from?

jaboutboul:
> I've blocked countries that are either frequent attackers to my servers, or run by dictators (ie: Russia, China, North Korea, etc). What country where you not able to access it from?

Hey @joacimmelin. We totally understand your desire to secure your mirror. However, without speaking to the political aspect of all this, the simplest *technical* issue is that every user in one of the blocked countries will have a timeout trying to determine the mirror’s speed when using the fastest mirror dnf plugin. We've actually tested this on our end and it adds close to 10 seconds of latency from some places. Have you considered the impact this will have?

joacimmelin:
I'll just remove the mirror then since I will not allow traffic from those countries.  Good luck with your project.

PR #88, opened 2021-04-02
Create Heficed mirrors 

Heficed3:
Hello,

can you please add our mirrors?

Thank you.


andrewlukoshko:
Hello and thanks for mirrors!
But files should be placed in mirrors.d/ not mirrors.d/repos/ so could you please update your pull request?

Heficed3:
Note that your request cannot be processed as it is submitted from an unregistered email address. To submit a ticket request, please send us a letter from the email address that you have used while purchasing our services or please login to 

If you have pre-sales questions please write an email to ***@***.***

PR #130, opened 2021-05-03
Added Cu.be Solutions mirror

wimg:
Currently trying to sync. Max connections reached, but will continue to try.


wimg:
Synchronized. Still having an issue doing the 3h rsync due to the connections limit, but the cron is there.

PR #131, opened 2021-05-04
First NZ mirror!

steriet:
Helping out from (middle) earth, New Zealand.


andrewlukoshko:
Wow! Thanks for mirror.
Waiting for sync.

steriet:
Iniital sync done, syncing every 3 hrs now.

PR #138, opened 2021-05-07
Added CloudCone.com Mirror

mr119:
Added [CloudCone](https://cloudcone.com) Mirror


andrewlukoshko:
Hello and thanks for mirror!
Please check your cron. Mirror is not updating.

mr119:
I think it should be fixed

PR #152, opened 2021-05-20
Update repo.dinosaurus.fleksibel.com.yml

ksatriabh:
add mail address


andrewlukoshko:
I've added email manually.
Thanks for update!

PR #157, opened 2021-05-24
Submit mirror.misakamikoto.network repo

Kwabang:
Location South Korea


andrewlukoshko:
Hi and thank you for mirror!
Could you please update mirror every 3-6 hours instead of 24h?
Currently we're marking mirrors with last sync more than 12h ago as outdated.

Kwabang:
> Hi and thank you for mirror!
> Could you please update mirror every 3-6 hours instead of 24h?
> Currently we're marking mirrors with last sync more than 12h ago as outdated.

Okay, I will update mirror every 6h.

PR #174, opened 2021-06-03
Update mirror.crexio.co.jp.yml

dokodemo5:
update frequency and added /


andrewlukoshko:
Frequency was updated on previous commit and "/" is not mandatory. Closing.

PR #176, opened 2021-06-06
Create almalinux.ipserverone.com.yml

yapwenbin:
New mirror - almalinux.ipserverone.com


yapwenbin:
Location: Kuala Lumpur, Malaysia

PR #189, opened 2021-06-24
Adding mirrors.d yaml file for Servers Australia mirror

ShaneHutter:
I've added an AlmaLinux mirror on the infrastructure for the company I work for, Servers Australia.  We are cPanel partners, and given that Alma will be supported by cPanel, and is maintained by Cloudlinux, a mirror on our network would likely be beneficial for both of us.


andrewlukoshko:
Thank you! Merged

PR #217, opened 2021-08-03
Added mirror.23m.com

vonlo123:


andrewlukoshko:
Hi! Thank for mirror!
Looks like it's not updating so could you please check your cron?

vonlo123:
Hi,

Thanks!
I had stupid mistake in the cron job.. It should sync every hour at minute 30.

Best regards 
Fabian von Lojewski

PR #248, opened 2021-09-21
create almalinux.antmediahost.com.yml

antmediahost:


andrewlukoshko:
Hello and thanks for mirror.
Please check your cron, looks like update frequency is not 1h, right now the mirror is outdated more than 12h.

antmediahost:
Hello,

Thank you..
We update cron to every 3h

PR #261, opened 2021-10-09
Table header style

srbala:
Table header missing CSS line

Signed-off-by: Bala Raman <srbala@gmail.com>


srbala:
seems fix needed on different branch

PR #264, opened 2021-10-11
Add UCSB to mirror list.

vwbusguy:
Signed-off-by: Scott Williams <scottwilliams@ucsb.edu>

Unfortunately, current refresh rate is 24h.  I've sent a request to change that internally and will update it if that happens.

### DIFF
--- a/mirrors.d/ftp.ucsb.edu.yml
+++ b/mirrors.d/ftp.ucsb.edu.yml
@@ -4,7 +4,7 @@ address:
   ftp: ftp://ftp.ucsb.edu/pub/mirrors/linux/almalinux/
   http: http://ftp.ucsb.edu/pub/mirrors/linux/almalinux/
   https: https://ftp.ucsb.edu/pub/mirrors/linux/almalinux/
-update_frequency: 24h
+update_frequency: 3h
 sponsor: University of California, Santa Barbara
 sponsor_url: https://www.ucsb.edu
 email: mtc@ucsb.edu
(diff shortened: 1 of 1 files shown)


jonathanspw:
Any luck getting the refresh rate changed?

vwbusguy:
Sadly no.  24 hours is the default here and Sys Admins with a different department maintain it.  I've requested that they update this, but that hasn't happened and might not?  Maybe just add us to the list for serving isos for now if that's too long for packages?

vwbusguy:
I sent a followup request about it today.

jonathanspw:
Great.

We don't have anything in place currently to let us only accept the mirror for ISOs while not serving it for updates.  The system would detect the mirror as outdated and not serve it at all.

vwbusguy:
I just heard back from the department and they advised that they've updated it to a 3 hour refresh rate, so I've updated the PR accordingly.

jonathanspw:
The syncing is more frequent but I'm not sure it's every 3h.  Right now the last sync was about 4.5 hours ago.  I'm going to keep an eye on it.

Can you please also add the geolocation fields per the [sample](https://github.com/AlmaLinux/mirrors/blob/master/mirrors.d/repo.almalinux.org.yml).

Thanks :)

vwbusguy:
I'll double check the sync time and add the geolocation stuff (I don't think that was in the sample when I originally submitted this PR).

jonathanspw (approved):

PR #267, opened 2021-10-21
Create atl.mirrors.knownhost.com

jonathanspw:
Add KnownHost Atlanta mirror


andrewlukoshko:
Looks like mirror is not updating, take a look https://atl.mirrors.knownhost.com/linux/almalinux/timestamp.txt

neonardo1:
Helps if I type the path right in the cron.  Fixed.

PR #270, opened 2021-10-24
Create almalinux-mirror.ereznet.co.il.yml

ereznet:


jonathanspw:
I see the mirror is online now but it is missing some of the core files in the main directory, having only the folders for each release.  Can you verify you setup the sync properly?

ereznet:
Please check our mirror:

http://mirrors.ereznet.co.il/almalinux/
https://mirrors.ereznet.co.il/almalinux/

Thank you !

PR #277, opened 2021-10-29
Create mirror.previder.nl.yml

j-insan3:


jonathanspw:
> Please add geolocation data per the sample at https://github.com/AlmaLinux/mirrors/blob/master/mirrors.d/repo.almalinux.org.yml

Went ahead and added it for you since the location was quite clear.

PR #279, opened 2021-10-29
Updated mirror.almalinux.si.yml

tomas-si:
Added geolocation information


jonathanspw:
The `state_province` value you summitted doesn't line up with our gelocation data (Nominatim).  I found what I believe it should be: https://nominatim.openstreetmap.org/ui/search.html?city=Domzale&state=Upravna+Enota+Dom%C5%BEale&country=SI

If this is correct can you please update your PR with a `state_province` value that will be recognized?

Also feel free to just further commit to https://github.com/tomas-si/mirrors/tree/tomas-si-patch-2 and it will automatically update this PR.  No need for new PRs ;)

PR #281, opened 2021-10-29
exclude mirrors which have no valid repos

jonathanspw:


src/backend/api/mirrors_update.py
```diff
@@ -245,6 +246,8 @@ async def mirror_available(
                     timeout=45,
                 ) as resp:
                     await resp.text()
+                    if resp.status == 200:
+                        has_valid_version_and_arch = True
```
soksanichenko, line 250: I think we can return just here result of checking

jonathanspw: Fair point.  If it has 1, it's valid.

PR #284, opened 2021-11-02
ALBS-51: Mirror service: implement private mirrors support

soksanichenko:
- The subject is implemented:
  - The json schema of a mirror config is updated.
   It's added field `private` which depends on field `subnets`.
   This field should be mandatory if `private` is set.
  - DB & data modules are updated. Also, it's added field `private`
   and the alembic migration for it
  - A private mirror shouldn't be checked because it can be behind NAT
  - [Fixed] Incorrect db query filter in
   function `backend.api.handlers._get_nearest_mirrors_by_network_data`


src/backend/api/mirrors_update.py
```diff
@@ -421,7 +432,7 @@ async def update_mirror_in_db(
     )
     db_session.add(mirror_to_create)
     logger.debug(
-        'Mirror "%s" is addded',
+        'Mirror "%s" is added',
```
rpisarev, line 435: What about formatting via f-string?

rpisarev, line 435: What about formatting via f-string?

rpisarev (approved):
lgtm

src/backend/api/mirrors_update.py
soksanichenko: Never don't use formatting/f-string for logging, because
1. It's slow way
2. logging has itself formatting
3. Sentry won't group events by a log message, because each of them will be unique due to a formatted variable

anfimovdm (approved):

PR #286, opened 2021-11-02
fix some asyncio blocking

jonathanspw:
Fixes some async blocking from `socket.gethostbyname()` and `requests`.

Converts to using `asyncio.Semaphores` for concurrency management.

### DIFF
--- a/src/backend/api/mirrors_update.py
+++ b/src/backend/api/mirrors_update.py
@@ -447,11 +447,12 @@ async def update_mirror_in_db(
 
 async def set_geo_data(
         mirror_info: MirrorYamlData,
-        sem,
+        sem: asyncio.Semaphore,
 ) -> MirrorData:
     """
     Set geo data by IP of a mirror
     :param mirror_info: Dict with info about a mirror
+    :param sem: asyncio Semaphore
     """
     mirror_name = mirror_info.name
     try:
--- a/src/backend/api/mirrors_update.py
+++ b/src/backend/api/mirrors_update.py
@@ -1,12 +1,12 @@
 #!/usr/bin/env python3
 import asyncio
+import aiodns
 import os
 from dataclasses import asdict
 
 import requests
 import yaml
 import dateparser
-import socket
 import time
 
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


src/backend/api/mirrors_update.py
```diff
@@ -249,17 +249,19 @@ async def mirror_available(
...
+                        # if mirror has no valid version/arch combos it is dead
+                        logger.error(
+                            'Mirror "%s" has one or more invalid repositories',
+                            mirror_name
                         )
-                        return mirror_name, True
-            except (ClientError, asyncio.TimeoutError) as err:
+                        return mirror_name, False
+            except (ClientError, asyncio.exceptions.TimeoutError) as err:
+                if type(err) == asyncio.exceptions.TimeoutError:
```
soksanichenko, line 263: 1. Please, use `isinstance(err, asyncio.exception.TimeoutError)`
2. And import `from asyncio.exceptions import TimeoutError` is better because it allows to use shorter the line here

src/backend/api/mirrors_update.py
```diff
@@ -249,17 +249,19 @@ async def mirror_available(
...
+                    err = type(err)
```
soksanichenko, line 264: For what is this conversion? Why do we not use variable `err` as is?

src/backend/api/mirrors_update.py
```diff
@@ -269,18 +271,19 @@ async def mirror_available(
                     err,
                 )
                 return mirror_name, False
-    # if mirror has no valid version/arch combos it is dead
-    logger.error(
-        'Mirror "%s" has no valid repositories',
-        mirror_name
+    # if mirror has at least one valid version/arch combo
```
soksanichenko, line 274: We loop through all versions and arches and we think a mirror is valid when it has all version/arches.

src/backend/api/mirrors_update.py
```diff
@@ -304,11 +307,13 @@ def set_repo_status(
...
     )
     try:
-        request = requests.get(
-            url=timestamp_url,
+        async with http_session.get(
+            timestamp_url,
             headers=HEADERS,
-        )
-        request.raise_for_status()
+            timeout=30,
```
soksanichenko, line 313: I suggest to make the number as module global constant and use it everywhere

src/backend/api/mirrors_update.py
```diff
@@ -437,19 +445,22 @@ async def update_mirror_in_db(
     )
 
 
-def set_geo_data(
+async def set_geo_data(
         mirror_info: MirrorYamlData,
+        sem,
```
soksanichenko, line 450: please add typing for the sem

soksanichenko (changes requested):

src/backend/api/mirrors_update.py
jonathanspw: Fix.  I tend to prefer longer method usage versus more specific imports for clarity when looking at code in the future, but I changed this as you wanted.

src/backend/api/mirrors_update.py
jonathanspw: Because `TimeoutError` has a blank error body so it simply results in something like this in the logs:

> Mirror "mirror.xyz.com" is not available for version "8.5" and repo path "PowerTools/debug/aarch64/" because ""

Instead of the blank reason now we get a reasonable entry such as:

> <class 'asyncio.exceptions.TimeoutError'>

in the "because" field so it is at least clear what happened.

src/backend/api/mirrors_update.py
jonathanspw: Fixed

src/backend/api/mirrors_update.py
jonathanspw: Fixed

soksanichenko (approved):

PR #287, opened 2021-11-02
aioredis, Nominatim cache

jonathanspw:
Updates redis to use aioredis w/hiredis.

Implement caching for Nominatim data via redis.

### DIFF
--- a/src/backend/api/redis.py
+++ b/src/backend/api/redis.py
@@ -5,6 +5,8 @@ from typing import (
     AnyStr,
     Optional,
     List,
+    Tuple,
+    Dict
 )
 
 from db.db_engine import RedisEngine
@@ -21,15 +23,15 @@ logger = get_logger(__name__)
 CACHE_EXPIRED_TIME = 24 * 3600  # 24 hours
 
 
-def get_mirrors_from_cache(
+async def get_mirrors_from_cache(
         key: AnyStr,
 ) -> Optional[List[MirrorData]]:
     """
     Get a cached list of mirrors for specified IP
     """
     key = str(key)
-    redis_engine = RedisEngine.get_instance()
-    mirrors_string = redis_engine.get(key)
+    redis_engine = await RedisEngine.get_instance()
+    mirrors_string = await redis_engine.get(key)
     if mirrors_string is not None:
         mirrors_json = json.loads(
             mirrors_string,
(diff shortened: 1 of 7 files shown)


src/backend/db/db_engine.py
```diff
@@ -72,12 +72,4 @@ class RedisEngine:
...
-        if not cls.__instance:
-            cls.__instance = Redis(
-                host=REDIS_HOST,
-                port=REDIS_PORT,
-                db=REDIS_DB,
-                charset="utf-8",
-                decode_responses=True
-            )
-        return cls.__instance
+        return aioredis.from_url("redis://%s:%s/%s" % (REDIS_HOST, REDIS_PORT, REDIS_DB))
```
soksanichenko, line 75: f-string is better variant and shorter than current

src/backend/api/redis.py
```diff
@@ -21,15 +23,17 @@
...
         key: AnyStr,
 ) -> Optional[List[MirrorData]]:
     """
     Get a cached list of mirrors for specified IP
     """
     key = str(key)
     redis_engine = RedisEngine.get_instance()
-    mirrors_string = redis_engine.get(key)
+    mirrors_string = await redis_engine.get(key)
+    await redis_engine.close()
```
soksanichenko, line 35: Can we move the two lines to a separate function?

src/backend/api/redis.py
```diff
@@ -48,20 +52,66 @@ def set_mirrors_to_cache(
...
+    key = str(key)
+    redis_engine = RedisEngine.get_instance()
+    coords = await redis_engine.get(key)
+    await redis_engine.close()
+    await redis_engine.connection_pool.disconnect()
+    if coords:
+        return json.loads(coords)
+
+
+async def set_geolocation_to_cache(key: AnyStr, coords: Tuple) -> None:
```
soksanichenko, line 74: FYI: python3.9 supports built-in types for typing)

soksanichenko (changes requested):

src/backend/api/redis.py
jonathanspw: The engine closing?  I don't see why...

src/backend/api/redis.py
jonathanspw: Thanks :)

soksanichenko (approved):

PR #290, opened 2021-11-04
Fix name of mirror.team-cymru.com

cschug:
The mirror `mirror.team-cymru.com` is not identical to `repo.almalinux.org` and therefore shouldn't be named like that.

### DIFF
--- a/mirrors.d/mirror.team-cymru.com.yml
+++ b/mirrors.d/mirror.team-cymru.com.yml
@@ -1,5 +1,5 @@
 ---
-name: repo.almalinux.org
+name: mirror.team-cymru.com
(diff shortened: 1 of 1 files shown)


jonathanspw:
Can you please also add geolocation data?

cschug:
@jonathanspw nope, I'm not operating that mirror and cannot give any details about it. I'm just a regular user of AlmaLinux scrolling thru the mirror list and was wondering why `repo.almalinux.org` was mentioned multiple times.

jonathanspw:
I see.  Good catch.  I'll do some digging and get geolocation added then.

PR #291, opened 2021-11-05
Mirror monitor and template updates (+ more)

jonathanspw:
- Change to aioredis lib w/hiredis
- Update Sentry SDK
- Enable Sentry performance tracking
- Enable Sentry aiohttp integration
- Disable Sentry redis integration (doesn't support aioredis yet)
- Update mirror monitoring interval to 1h
- Track mirror flapping.
  - A mirror is marked as flapping when it has failed a health check within the past 12 hours.  Mirrors marked as flapping are still displayed on the mirrors page but are not served as part of `/mirrorlist` requests.
- Improve DB call efficiency for listing all mirrors.  The result is the page generation time is now under 50ms on average instead of 350ms+
- Improve DB call efficiency for mirrorlist requests.  The result is the generation time is on average 50-100ms faster than before.
- Add tracking of which mirrors support IPv6
- Add generic favicon to prevent browsers from hitting 404s when checking default URLs for `/favicon.ico`


jonathanspw:
image

src/backend/db/models.py
```diff
@@ -128,7 +128,7 @@ class Mirror(Base):
     ip = Column(String, nullable=False)
     latitude = Column(Float, nullable=False)
     longitude = Column(Float, nullable=False)
-    is_expired = Column(Boolean, nullable=False, default=False)
+    status = Column(String, nullable=False)
```
soksanichenko, line 131: Did you make db migrations?

src/backend/templates/mirrors.html
```diff
@@ -94,6 +94,11 @@
                             {% endif %}
                         </td>
                     {% endfor %}
+                    <td>
+                        {% if mirror.ipv6 %}
```
soksanichenko, line 98: Did you add the column to header?

soksanichenko (changes requested):

src/backend/db/models.py
jonathanspw: sqlite cannot change column type so no I didn't.  Do we want one that simply removes the is_expired col and adds status?

src/backend/templates/mirrors.html
jonathanspw: Yep https://github.com/AlmaLinux/mirrors/pull/291/commits/983abaf6198191a924e5546e6e6b29393c3df0b8#diff-7dc0ea0263421b070b14c289b52f9467996fb61796dbd1277b1c99ea1da14e4fR143

src/backend/db/models.py
soksanichenko: I did migration to postgres which can do it)

soksanichenko: Never mind. I just mix up services.

soksanichenko (approved):

PR #292, opened 2021-11-08
Cache mirrorlist in redis

jonathanspw:
Mostly just code to cache the mirrorlist in redis to avoid slow DB calls as much as possible.

- ISOs page load time went from >300ms to ~80ms
- `/mirrorlist/*` load time from >300ms to ~80ms
- Main HTML mirror page sped up a tad more
- Lower mirrorlist caching per IP to 1h from 24h.

This results in significant CPU savings/higher TPM.

`/mirrorlist/*` can now serve about 25 req/s on pure uncached requests using our standard AWS instance size for mirror service servers.  On cached requests we can do 115 req/s per instance.

Logic related to subnets is the most intensive part of the whole system now but I can't really come up with a way to improve the efficiency on that (yet).


soksanichenko (commented):
The PR should be rebased and merged after fixing previous PRs

soksanichenko (approved):

PR #307, opened 2021-11-09
updating the Continent

sysdev-isu:
the Continent is unknown in https://mirrors.almalinux.org/


jonathanspw:
Continent is not a valid field.  We rely on geoip for that.

PR #312, opened 2021-11-09
add geolocation for ftp.fau.de

rrzefox:
this commit adds the correct geolocation for ftp.fau.de.

Note: the geolocation in repo.almalinux.org.yml (which was used as a template to copy from) seems to be wrong: repo.almalinux.org seems to be hosted with Hetzner, but their data center is in Falkenstein/Vogtland, NOT the Falkenstein in Bavaria. So the correct state_province for repo.almalinux.org would be Saxony, NOT Bavaria!


jonathanspw:
Thanks for the report, I'm fixing that in https://github.com/AlmaLinux/mirrors/pull/337

PR #323, opened 2021-11-09
Adding phx.mirrors.clouvider.net

MaciejKupiec:
Add mirror for Phoenix


mirrors.d/phx.mirrors.clouvider.net.yml
```diff
@@ -0,0 +1,10 @@
+address:
```
jonathanspw, line 1: You forgot the name field.  Please add it.

jonathanspw (changes requested):

PR #348, opened 2021-11-12
Add geolocation information for almalinux.uib.no

stompa75:


jonathanspw:
I'm not sure why this is failing CI even though it looks valid.  I'll need to do some debugging on that last CI test to make sure everything is good before merging this.

PR #354, opened 2021-11-15
ALBS-74: Some improvements for the mirrors service

soksanichenko:
- [Improvement] The mirrors from request's country should be first in the list
- [Improvement] Downgrade the level of logging of issue "Error retrieving Nominatim data for..." to warning
- Refactoring
  - Make context manager for Redis async engine
  - Some logged events should be logged as exceptions not errors
  - Typing according to version of Python (3.9)
  - Docstrings


src/backend/api/utils.py
```diff
@@ -185,10 +187,9 @@ async def get_azure_subnets_json(http_session: ClientSession) -> Optional[Dict]:
             response_text = await resp.text()
     except client_exceptions.ClientConnectorError as err:
         logger.error(
```
anfimovdm, line 189: May using `exception` method be better than `error` when catching exception?

anfimovdm (changes requested):

src/backend/api/utils.py
soksanichenko: Done

anfimovdm (approved):

Korulag (approved):

anfimovdm (approved):

ci/ansible/roles/deploy/defaults/main/packages.yml
```diff
@@ -0,0 +1,37 @@
...
+  - geoip2==4.4.0
+  - haversine==2.5.1
+  - jsonschema==3.2.0
+  - Jinja2==3.0.2
+  - lxml==4.6.3
+  - pyyaml==6.0
+  - requests==2.26.0
+  - sentry-sdk[flask]==1.4.3
+  - sqlalchemy==1.4.26
+  - uwsgi==2.0.20
```
Korulag, line 36: Mirror service still using uwsgi?

ci/ansible/roles/deploy/defaults/main/packages.yml
```diff
@@ -0,0 +1,37 @@
...
+  - Flask-API==3.0.post1
+  - geoip2==4.4.0
+  - haversine==2.5.1
+  - jsonschema==3.2.0
+  - Jinja2==3.0.2
+  - lxml==4.6.3
+  - pyyaml==6.0
+  - requests==2.26.0
+  - sentry-sdk[flask]==1.4.3
+  - sqlalchemy==1.4.26
```
Korulag, line 35: 1.4.27 already available

soksanichenko: Done

ci/ansible/roles/deploy/defaults/main/packages.yml
soksanichenko: Yep. I'll replace it by gunicorn when I migrate from SQLite to PosgreSQL

PR #356, opened 2021-11-15
add mirror.karneval.cz

xjoska:


xjoska:
Hello,

do you have any idea how can be fixed failed nominatim-check ?

geolocation:
  country: CZ
  state_province:
  city: Prague

It seems to be valid at https://nominatim.openstreetmap.org/search?city=Prague&state=&country=CZ&format=json

PR #372, opened 2021-11-24
ALBS-75: Make CI on GH for the mirror service

soksanichenko:
- It's added the submodule
- CI of checking the mirrors configs and the main config


soksanichenko:
wrong

PR #374, opened 2021-11-25
Update mirror.dewabiz.com.yml

dewabiz20:


jonathanspw:
Can you put the sync frequency back at 3h or less?

jonathanspw:
The Nominatim location also needs to be fixed.

dewabiz20:
Hi, the sync frequency is set back to every 3h, and nomination location also set,

PR #377, opened 2021-11-26
Update ftp.lug.ro.yml

cmatei:
There is no state/province administrative division in Romania, just 41 counties plus Bucharest. Using Bucharest in the state field on nominatim yields 0 results, so I ommited the state_province field in the mirror template. If that's not ok, can you please suggest what to use instead?


jonathanspw:
The field should be present but blank.

PR #384, opened 2021-11-29
Update almalinux.mirrors.as212647.net.yml

mikespears:
I've fixed this mirror, please re-add to the list


jonathanspw:
@mikespears please make one more PR to add geolocation data. I accidentally merged this without it being there.

PR #386, opened 2021-11-30
added FlokiNET mirror

zehkaFlokinet:
An almalinux mirror ist available under mirror.flokinet.net/almalinux via http, https and rsync

### DIFF
--- a/mirrors.d/mirror.flokinet.net.yml
+++ b/mirrors.d/mirror.flokinet.net.yml
@@ -1,0 +1,15 @@
+---
+name: mirror.flokinet.net
+address:
+  http: http://mirror.flokinet.net/almalinux/
+  https: https://mirror.flokinet.net/almalinux/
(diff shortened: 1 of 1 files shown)


soksanichenko:
I guess the value of state should be quoted

zehkaFlokinet:
If that's the case that should have fixed it

soksanichenko:
I'll check PR manually tomorrow

jonathanspw:
Quoting isn't necessary.  YAML recognizes spaces within values.

In this case the `state_province` field should be left blank as Nominatim has no recognized state-level boundary for Bucharest.

https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=377733&class=boundary

soksanichenko:
@jonathanspw thank you. I should add more detailed log for this check 🤔

PR #392, opened 2021-12-02
New mirror for AlmaLinux sponsored by Access Communications

maxglaz:
New mirror for AlmaLinux sponsored by Access Communications.
Located at Access Communications DC at Regina.
Multiple 100G connections to Canadian ISPs.


jonathanspw:
Hi,

It looks like your mirror has dropped offline.  Will it be fixed or is it permanently offline?

PR #396, opened 2021-12-11
Deleting 3 mirror sites

trippyboy:
I would like to stop serving those mirror sites due to the increase of required disk space which has been increased since the mirror guide has release in the first place. And I see a lot of mirror sites are there, so I would like to close those.

Thank you,
trippyboy


jonathanspw:
Sorry to see you taking your mirrors down.  If you have the capacity in the future we'd love to have you back!

PR #408, opened 2021-12-29
Create repo.inara.pk.almalinux.yml

ahrar25:
Alma Linux Repository mirror in Pakistan


jonathanspw:
Hi thanks for your submission.  Please double check the geo data as it is incorrect.

ahrar25 (commented):
I have changed state_province to Islamabad and city to Islamabad as well, Since it is the capital it does not fall in any province. Islamabad goes for state and city both. please do approve

PR #412, opened 2022-01-10
Create muug.ca.yml

athompso:
If there's a way to indicate we support HTTP, HTTPS, FTP, and RSYNC that would be nice... and also the fact we're connected at 10Gbps.  (Although for \<reasons\>, we can only serve files up at ~8.5Gbps max.)

(Technically I'm jumping the gun, the initial rsync is still going.  I'm assuming it'll be done by the time someone gets around to reviewing this.)

### DIFF
--- a/mirrors.d/muug.ca.yml
+++ b/mirrors.d/muug.ca.yml
@@ -1,0 +1,15 @@
+---
+name: muug.ca
+address:
+  http: http://muug.ca/almalinux/
+  https: https://muug.ca/almalinux/
+  rsync: rsync://muug.ca/almalinux
+geolocation:
+  country: CA
+  state_province: Manitoba
+  city: Winnipeg
+update_frequency: 1h
+sponsor: Manitoba UNIX User Group
+sponsor_url: http://muug.ca/
+email: mirror@muug.ca 
+...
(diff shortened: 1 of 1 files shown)


jonathanspw:
> If there's a way to indicate we support HTTP, HTTPS, FTP, and RSYNC that would be nice

That's covered by simply providing the URLs in the config that you support.

jonathanspw:
Your URLs all 404.

athompso:
Argh.  I see what I did there... fixed.  Sorry about that.  (To be clear, the rsync URL was always working... which is relevant to about 0.01% of users.  If you got failures on that one, we need to figure out why!)

Yes, I see the HTTP/HTTPS/RSYNC/IPv6 columns there, not sure what I was thinking when I wrote that comment.  Although speed would be a nice thing to show on that page (i.e. 1/10/100G)

athompso:
Also, I didn't see any requirement for this, but if you need specific public paths exposed (e.g. OpenBSD must always live at /pub/OpenBSD) we can do that.

athompso:
I've manually checked the URLs, what did I miss this time???

jonathanspw:
I'm not sure.  I'll do some digging and try to see what it's failing on specifically.

athompso:
Finally got time to go through config_checker.py, and I suspect it might be barfing on http_session not being able to handle the rsync protocol.  Not fluent enough in Python to be certain of that by any means, though.

PR #419, opened 2022-01-18
Install Nginx's from the official mainline repository

LKHN:
Install a more up-to-date version of Nginx from its mainline RPM repository instead of AppStream, which is marked as a "Legacy" version (1.14)

Signed-off-by: Elkhan Mammadli <elkhan.mammadli@protonmail.com>


jonathanspw:
Latest VM images have been upgraed to Alma 9 and have nginx 1.20.  Unless there's a specific benefit in a version from the mainline repository I don't think we should do this.

PR #420, opened 2022-01-19
ALBS-133: [ASAP] Fix growing LA on the mirrors service

soksanichenko:
- The service is migrated from uWSGI to Gunicorn
  - It's created SELinux rules for Gunicorn
- [Fixed] We always get IP for private mirrors
- [Fixed] We check beta versions for Azure mirrors


ci/ansible/roles/deploy/templates/gunicorn.py.j2
```diff
@@ -0,0 +1,24 @@
+wsgi_app = 'main:app'
+bind = '[::1]:{{ uwsgi_port }}'
```
Korulag, line 2: Maybe rename to `gunicorn_port`?

PR #431, opened 2022-02-03
ALBS-139: The service should be able to return only a private mirror for suitable clients

soksanichenko:
- Upload to a server yaml_snippets
- Wait for update the mirrors list before run ansible tests of the service
- Pass variable `source_path` to the backend environment
- Make pidfile when updating of the mirrors list is running and remove pidfile when it's ended
- Description of backend systemd service is fixed
- Version of JSON schema of a mirror config is increased
- Field `monopoly` is added to dataclass MirrorData
- DB migration for the new field is created
- It's removed duplicated modules: data_models.py & json_schemas.py
- Imports in a lot of modules are fixed
- The service returns only one private mirrors if it's
  suitable for client and the mirror is monopoly
- Some fixes in comments
- It's removed duplicated functions from mirrors_update.py
- Logging level of some messages from api.utils.py is reduced from exception to error
- It's fixed the issue with empty list AWS/Azure subnets


src/backend/api/handlers.py
```diff
@@ -213,14 +213,29 @@ async def _process_mirror(
...
         os.getenv('CONFIG_ROOT'),
         'mirrors/updates',
         config.mirrors_dir,
     )
     all_mirrors = get_mirrors_info(
         mirrors_dir=mirrors_dir,
+        logger=logger,
+        path_to_json_schema=os.path.join(
+            os.environ['SOURCE_PATH'],
+            'src/backend/yaml_snippets/json_schemas/mirror_config.json',
```
rpisarev, line 237: Well, I saw the constant 4 times. If I were you I would use the constant instead of literal

rpisarev (approved):

monkrobot (approved):

Korulag (approved):

anfimovdm (approved):

PR #461, opened 2022-03-05
Create mirror.parsdev.com.yml

parsdev-com:


parsdev-com:
@jonathanspw Hello why not listed our mirror on https://mirrors.almalinux.org/ ?

PR #462, opened 2022-03-10
New mirror in Greece

bilias:
Initial sync in progress.

Could be nice if repo.almalinux.org provided rsync access as well. Mirror site lists rsync://rsync.repo.almalinux.org/almalinux which is far/slow from EU


jonathanspw:
You can do your initial sync from any mirror so it's faster, just use rsync.repo.almalinux.org in your cron job.

bilias:
Yes I did use another one to sync.
sync is done

PR #468, opened 2022-03-21
ALBS-228: Fix error `UnicodeDecodeError`

soksanichenko:


Korulag (approved):

utils.py
```diff
@@ -354,7 +354,15 @@ async def mirror_available(
                         headers=HEADERS,
                         timeout=AIOHTTP_TIMEOUT,
                 ) as resp:
-                    await resp.text()
+                    try:
+                        await resp.text()
+                    except Exception as err:
+                        logger.error(
```
anfimovdm, line 360: You can use `logger.exception` instead of `logger.error`, same in #467

anfimovdm (changes requested):

PR #492, opened 2022-04-22
Update almalinux.mirror.myduniahost.com.yml

bedanish11:
Change http, rsync and sponsor


jonathanspw:
Nothing changed in this.

PR #497, opened 2022-04-25
Create mirror.0-1.cloud.yml

MrAriaNet:
Please added my mirror in list


jonathanspw:
Looks like your mirror is offline (AlmaLinux has been removed from it).  Will it be coming back?

PR #498, opened 2022-04-25
Add mirrors.kernel.org entries

mricon:
There are multiple nodes serving mirrors.kernel.org, with one of them
only serving rsync, and others only serving https. Normally, they are
accessed as mirrors.edge.kernel.org with geoDNS magic routing the
request to the nearest node, but it's possible to localize down to a
specific geographical area.

Signed-off-by: Konstantin Ryabitsev <konstantin@linuxfoundation.org>

### DIFF
--- a/mirrors.d/ap.edge.kernel.org.yml
+++ b/mirrors.d/ap.edge.kernel.org.yml
@@ -1,0 +1,12 @@
+---
+name: ap.edge.kernel.org
+address:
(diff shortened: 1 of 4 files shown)


jonathanspw:
Thanks for setting up the mirrors.  Looks like CI is having an issue related to some reason functionality change in our mirror service so we'll get that sorted and get your mirrors merged shortly.

PR #500, opened 2022-04-26
ALBS-293: Mirror service: Add config option to set repo as available only on vault_mirror

soksanichenko:
- The JSON schema of the service config, loading the service config and the data model
  are changed according to implementation of the subject
- The separate function checks that a repo is vault or not

### DIFF
--- a/src/backend/api/handlers.py
+++ b/src/backend/api/handlers.py
@@ -354,6 +354,29 @@ async def get_all_mirrors_db(
     return mirrors_list
 
 
+def _is_vault_repo(
+        version: str,
+        vault_versions: list[str],
+        repo: RepoData,
+) -> bool:
+    """
+    Check that the repo is vault or not.
+      The function returns True if repo is vault and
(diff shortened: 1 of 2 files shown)


src/backend/api/handlers.py
```diff
@@ -354,6 +354,27 @@ async def get_all_mirrors_db(
...
+    """
+    Check that the repo is vault or not.
+      The function returns True if repo is vault and
+      returns False if the one isn't vault
+    :param version: version of requested a mirrors list
+    :param vault_versions: the list of global vault versions
+    :param repo: repo of requested a mirrors list
+    """
+
+    if version in vault_versions:
```
anfimovdm, line 371: You can union this conditions

anfimovdm (approved):

src/backend/api/handlers.py
soksanichenko: Done

PR #516, opened 2022-05-03
Adding back codingflyboy.mm.fcix.net after XSLT fix

PhirePhly:
The root cause for this mirror mangling the repomd.xml files was that
the xslt styling for directory indexes was being applied to "location /"
in Nginx, where the correct location to apply it was "location ~ /$" to
only apply the templating to paths ending in / such that there is no
way that actual assets could be modified.

Relevant commit in our playbook:
https://github.com/PhirePhly/micromirrors/commit/9057f49792ae3a648206910f320dbba7fc98bd35

I apologize for causing problems the first time.

### DIFF
--- a/mirrors.d/codingflyboy.mm.fcix.net.yml
+++ b/mirrors.d/codingflyboy.mm.fcix.net.yml
@@ -1,0 +1,18 @@
+---
+name: codingflyboy.mm.fcix.net
+address:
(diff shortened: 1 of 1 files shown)


PhirePhly:
I will also point out that the revert #511 didn't seem to actually completely disable this mirror. We were still visible on the ISO download page, and I was still seeing some hosts requesting updates from our repo.

I'm not sure if that was expected behavior or not...

PR #582, opened 2022-07-13
Added ftp.belnet.be to the mirrors

pepouille:


jonathanspw:
Mirror is completely offline it seems.

PR #583, opened 2022-07-15
Mirrors service updates

aronowski:


ci/ansible/roles/deploy/templates/nginx.conf.j2
```diff
@@ -52,6 +52,11 @@ http {
                     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
                 }
 
+                location /mirrorlist {
```
soksanichenko, line 55: For what is it necessary?
I asked because `vuepress` part is not used in the project long ago

soksanichenko (changes requested):

ci/ansible/roles/deploy/templates/nginx.conf.j2
aronowski: According to the knowledge gained from forking the project, for my configuration this was required for the mirrorlist to be actually served properly. Without this, I was getting an empty response.

A nice addition is also the fact that one can browse the mirrorlist via a browser this way. Compare the following:
https://mirrors.almalinux.org/mirrorlist/9/
http://mirrors.cdn.euro-linux.com/mirrorlist/9/

soksanichenko: > Without this, I was getting an empty response.
What does endpoint return empty response?
> A nice addition is also the fact that one can browse the mirrorlist via a browser this way. 
1. We designed endpoint `https://mirrors.almalinux.org/mirrorlist/<version>/<repo>` specially for DNF/YUM :)
2. We can add some `static` page with sublevels, but it will also be generated on backend side. The page won't use `vuepress`

PR #588, opened 2022-07-31
Update FCIX Mirror Fleet

PhirePhly:
Thanks for the list of ASN support! That allowed me to more easily add all the networks on the FCIX exchange which specifically wanted to use that mirror.

Also adding new 10G mirror in Virginia.


jonathanspw:
I'm not sure this is the intended use of the ASN lists.  This is going to make all of those ASNs hit mirror.fcix.net which geographically is not likely to be ideal.

PhirePhly:
This is the intention. We didn't include all 140 ASNs on FCIX, just the ones who specifically requested to be included in the traffic engineering for mirror.fcix.net

warthog9:
It might be helpful to note the the respective mirrors noted in the PR the ASNs in question are in fact directly peered with the IX where the mirror is at:

https://fcix.net/participants/

might be helpful.  As noted by @PhirePhly we explicitly asked the IX participants if they wanted to be listed thusly, and we are already doing this with Fedora and EPEL with this list.

warthog9:
Also as AS53758 I can confirm this is expected for that ASN

PR #639, opened 2022-11-10
Added ziply.mm.fcix.net

PhirePhly:
I'm not deliberately doing these all as individual pull requests. I just wasn't expecting this node for another week. -_-


jonathanspw:
We don't mind the PRs :)

PR #650, opened 2022-11-12
Add ftp.sh.cvut.cz

pdostal:
Hello,

I'd like to add `ftp.sh.cvut.cz` to the mirror list.
 * We're in the Czech academical network - CESNET.
 * Our mirror is connected via 10gbps fibre.

### DIFF
--- a/mirrors.d/ftp.sh.cvut.cz.yml
+++ b/mirrors.d/ftp.sh.cvut.cz.yml
@@ -1,0 +1,19 @@
+---
+name: ftp.sh.cvut.cz
+address:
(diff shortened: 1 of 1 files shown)


pdostal:
Should I add all the IP ranges within our network or is the AS number enough for your mirror brain?

jonathanspw:
Adding subnets and/or ASN will cause all traffic within your ASN/Subnets to hit your mirror instead of reaching out to others.

Generally one or the other is all that's needed.

tuxjz (approved):
Looks good 👍

PR #660, opened 2022-11-15
ALBS-396 Add some check of existing ISOs on an AL mirror

soksanichenko:
- Unified name of services/scripts
- A few utility packages for installation
- PID mechanism for avoiding duplicating of update
- Some refactoring


src/backend/api/handlers.py
```diff
@@ -255,49 +260,62 @@ async def update_mirrors_handler() -> str:
...
+                            mirror_info=mirror_info,
+                            db_session=db_session,
+                            http_session=http_session,
+                            nominatim_sem=nominatim_sem,
+                            mirror_check_sem=mirror_check_sem,
+                            main_config=config,
+                        )
+                    ) for mirror_info in all_mirrors
+                ))
+            db_session.flush()
```
anfimovdm, line 299: I think you don't need to use `flush` here, because your session is closing at this point

anfimovdm (approved):

src/backend/api/handlers.py
soksanichenko: Thanks. Done.

PR #665, opened 2022-11-17
ALBS-396: Add some check of existing ISOs on an AL mirror

soksanichenko:
- The subject is implemented
- List of the closest ISO mirrors will be cached to Redis


maccelf (approved):

src/backend/api/handlers.py
```diff
@@ -375,6 +401,10 @@ async def get_all_mirrors_db(
                     Mirror.private == null()
                 ),
             )
+        if iso_mirrors:
+            mirrors_query = mirrors_query.filter(
+                Mirror.has_full_iso_set == true(),
```
anfimovdm, line 406: you can use `field.is_(True)` instead

anfimovdm (approved):

Korulag (approved):

anfimovdm (approved):

src/backend/api/handlers.py
```diff
@@ -222,95 +211,164 @@ async def _get_nearest_mirrors(
...
+                    continue
+                if mirror_info.status == 'flapping':
+                    continue
+                urls_to_create = [
+                    Url(
+                        url=url,
+                        type=url_type,
+                    ) for url_type, url in mirror_info.urls.items()
+                ]
+                for url_to_create in urls_to_create:
```
Korulag, line 335: sqlalchemy has `add_all` to add a list of entities

src/backend/api/handlers.py
```diff
@@ -222,95 +211,164 @@ async def _get_nearest_mirrors(
...
+                    has_full_iso_set=mirror_info.has_full_iso_set,
+                )
+                if mirror_info.subnets:
+                    subnets_to_create = [
+                        Subnet(
+                            subnet=subnet,
+                        ) for subnet in mirror_info.subnets
+                    ]
+                    for subnet_to_create in subnets_to_create:
+                        db_session.add(subnet_to_create)
```
Korulag, line 369: Same as above

Korulag (approved):

PR #671, opened 2022-11-19
update continent of mirror.ossplanet.net

bluet:
East Asia -> Asia


jonathanspw:
East Asia is not a continent.

bluet:
@jonathanspw That is exactly the reason why I open this PR to change from "East Asia" to "Asia". Would you mind checking the "Files Changed" tab and have a look?

jonathanspw:
I read the PR backwards :)

PR #676, opened 2022-11-29
Update repo.almalinux.org.yml

nkondarev:
Hello, WWFX would like to join the ALMALINUX community

Cheers,


andrewlukoshko:
Hello and thanks for mirror. Please don’t update our mirror file.
You should provide separate file for your mirror.

PR #678, opened 2022-11-30
Update repo.almalinux.org.yml

nkondarev:


soksanichenko:
Hi @nkondarev
Thank you for your contribution! 
But you should put changes to a new separate config.

soksanichenko (changes requested):
It's necessary a separate config

PR #693, opened 2022-12-13
Tino Group update mirror

tranbinhcse:
Hello,
We have implemented a new mirror server,
Please check.


jonathanspw:
It would be great if you could re-do your changes on a rebased copy of the repo - it looks like you made this commit against a really old copy.

I'd also prefer if you did a "mv" or similar rename for the file and modified it as necessary instead of removing and recreating it - but that's just a preference.

PR #697, opened 2022-12-13
Create mirrors.ucr.ac.cr.yml

valarauco:
Adding public mirror hosted at Universidad de Costa Rica

### DIFF
--- a/mirrors.d/mirrors.ucr.ac.cr.yml
+++ b/mirrors.d/mirrors.ucr.ac.cr.yml
@@ -1,0 +1,19 @@
+---
+name: mirrors.ucr.ac.cr
+address:
+  http: http://mirrors.ucr.ac.cr/almalinux/
+  https: https://mirrors.ucr.ac.cr/almalinux/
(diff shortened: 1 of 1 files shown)


jonathanspw:
"Central America" is not a valid continent.  Can you please remove that line entirely (the system will pull it automatically based on the country) or change it to "South America"?

valarauco:
Central America should be valid. Sometimes they put us in North America and other times in the South. We are small but we exist. 🤣
According to Wikipedia we are part of North America, tho 😅

jonathanspw:
OpenStreetMap/Nominatim says South America https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=120027&class=boundary

Either way all will be fine with the line removed.  GeoIP will pull it appropriately.

jonathanspw:
Whoops, we're talking about Costa Rica not Columbia :)  So North for sure.

valarauco:
Hohoho
Relax, sometimes people also confuse us with Puerto Rico 👍

PR #701, opened 2022-12-14
Create mirrors.upr.edu.yml

selopez-upr:
Description for mirrors.upr.edu


jonathanspw:
Can you please double check your geo data against https://nominatim.openstreetmap.org/ui/details.html ?  What you have set currently is not valid.

PR #718, opened 2023-01-06
Add OSUOSL mirrors

ramereth:
I have submitted requests to fix Geo location of all systems as some of them were a little off.

Signed-off-by: Lance Albertson <lance@osuosl.org>


jonathanspw:
CI failed because the ASN has to be only the number - ie not prefixed with "AS".  Can you fix this please?

PR #725, opened 2023-01-19
Create almalinux.brontocdn.net.yml

brontocdn:
BrontoCDN Mirror pull request

### DIFF
--- a/mirrors.d/almalinux.brontocdn.net.yml
+++ b/mirrors.d/almalinux.brontocdn.net.yml
@@ -1,0 +1,14 @@
+---
+name: almalinux.brontocdn.net
+address:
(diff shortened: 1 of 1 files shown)


jonathanspw:
Thanks for the mirror submission.

As I can see your site is full of default "lorem ipsum" content and your domain was only registered 1 week ago.  Are you sure that you have the capacity and network infrastructure to handle a mirror?  I believe you're running this on an OVH dedicated server, correct?  It's quite likely mirroring would send you over bandwidth limits.

brontocdn:
Hi Jonathan, 

In Germany, we have a mirror pool. And the total storage capacity is 15 TB with 5 Gbps outgoing traffic. I believe it will suffice.

PR #746, opened 2023-02-27
update mirrors.isu.net.sa.yml

sysdev-isu:
adding rsync link


sysdev-isu (commented):
Dears,

The mirror is not listed on mirrors.almalinux.org

Regards

PR #801, opened 2023-05-17
Add mirror.pulsant.com.yml

thomascurrie:
Adding the Pulsant mirror to the Almalinux mirror list


jonathanspw:
It looks like your mirror has dropped offline.  Will this be fixed or is the mirror permanently offline?

PR #807, opened 2023-05-27
Transfer to a new domain (mirror.yuki.net.uk)

Kwabang:
I will add mirror.yuki.net.uk with a new PR.


Kwabang:
CI is having problems by deleting files from the mirror list.

PR #819, opened 2023-06-19
moved mirror.flokinet.net to location specific subdomain

zehkaFlokinet:
We are currently rolling out new mirrors in different countries so we assigned our pre-existing mirror a location specific prefix. The old mirror.flokinet.net domain will remain valid but will at some point be converted to an anycast ip that routes to our closest by mirror


jonathanspw:
> We are currently rolling out new mirrors in different countries so we assigned our pre-existing mirror a location specific prefix. The old mirror.flokinet.net domain will remain valid but will at some point be converted to an anycast ip that routes to our closest by mirror

Please maintain a config-per-mirror in git with the respective locations of each, even if it's anycast.

zehkaFlokinet:
That's what we are doing. In fact this pull request is about removing the generic "mirror.flokinet.net" from the mirrors list and add the first specific location (ro.mirror.flokinet.net) instead

PR #821, opened 2023-06-25
Update mirror.cenuta.com.yml

mertcenikut:
Update URL + Location

PR #822, opened 2023-06-25
Revert "Update mirror.cenuta.com.yml"

codyro:
Reverts AlmaLinux/mirrors#821


codyro:
After looking over this again, invalid `https` mirror (points to `https://mirror.centos.org`).

PR #828, opened 2023-06-30
alma.acidman.thelefty.org is now open to the world

metalefty:
While here, fix sponsor_url.

### DIFF
--- a/mirrors.d/alma.acidman.thelefty.org.yaml
+++ b/mirrors.d/alma.acidman.thelefty.org.yaml
@@ -9,13 +9,7 @@ geolocation:
   city: Fukuoka
 update_frequency: 3h
 sponsor: Monochrome Rainbow
-sponsor_url: http:/www.thelefty.org/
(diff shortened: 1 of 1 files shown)


jonathanspw:
Do you still want the previously-defined subnets and ASN to hit your mirror primarily?  If so, those sections should stay.

metalefty:
Yes, I'll put it back and add an IPv6 subnet.

jonathanspw:
Oh I missed the fact that you did leave the ASN.  If the subnets are within the defined ASN then there's no need for the subnets.

Also, thanks for making it public :)

metalefty:
Understood, thanks!

metalefty:
@jonathanspw Just curious, when will it appear on mirrors.almalinux.org?

codyro:
This should be appearing by now. 

I'll check once I'm on the PC, or @soksanichenko can shed some insight as well.

PR #851, opened 2023-08-10
Need to remove mirror server

matsuricorp:
Need to remove mirror server.


jonathanspw:
Sorry to see you take down the mirror, but thank you for the contribution for the time you had it!

PR #852, opened 2023-08-13
Replace mirror.host4.biz with mirror.host4biz.eu

jaredhared:


jonathanspw:
Can you please add the geolocation data to the config?

jaredhared:
> Can you please add the geolocation data to the config?
It was there, anyway, added state_province.

jonathanspw:
> > Can you please add the geolocation data to the config?
> 
> It was there, anyway, added state_province.

That let CI pass, thanks!

PR #862, opened 2023-08-24
Create aws.repo.almalinux.org.yml

jonathanspw:


mirrors.d/aws.repo.almalinux.org.yml
```diff
@@ -0,0 +1,16 @@
+---
+name: aws.repo.almalinux.org
+address:
+  http: http://aws.repo.almalinux.org/almalinux/
```
jonathanspw, line 4:
```suggestion
  http: http://aws.repo.almalinux.org/almalinux/
  https: https://aws.repo.almalinux.org/almalinux/
```

codyro:
Should I be getting a `403` when attempting to view/download things in `/iso`?

This output is from an instance in `us-east-1`:

```bash
[ec2-user@bedtimeyet ~]$ curl -I https://aws.repo.almalinux.org/almalinux/9.2/isos/x86_64/AlmaLinux-9.2-x86_64-dvd.iso
HTTP/2 403
content-type: text/html
content-length: 153
server: nginx/1.20.1
date: Thu, 24 Aug 2023 19:19:55 GMT
x-cache: Error from cloudfront
via: 1.1 bd5ab138e49a090fd3089313c9c8105c.cloudfront.net (CloudFront)
x-amz-cf-pop: IAD12-P4
alt-svc: h3=":443"; ma=86400
x-amz-cf-id: 0PkbfHRCPyoXeKUm95wqYE3c7I3CZaLUkg5bKxHUVqbPU9uw4N3vdQ==
111
```

On the CloudFront side, it looks straightforward/sane.

jonathanspw:
Yes the 403 on isos is intentional.  We do not push them to Azure mirrors either as these are intended to be update-only mirrors for the given cloud (AWS in this case).

codyro:
Not that it matters, but we may as well do the same for `/live/` and (maybe?) `/raspberrypi/images/` for consistency.

codyro (approved):

jonathanspw:
> Not that it matters, but we may as well do the same for `/live/` and (maybe?) `/raspberrypi/images/` for consistency.

Fair point, I missed them.  Will adjust.

PR #863, opened 2023-08-25
fix/replace the @cache.cached() decorator

jonathanspw:


codyro (approved):
The logic makes sense to me.

What was the issue with `@cache.cached()` that required it to be replaced/worked around?

jonathanspw:
> The logic makes sense to me.
> 
> What was the issue with `@cache.cached()` that required it to be replaced/worked around?

Simply put, it wasn't working, at least not consistently.  It was randomly not caching certain mirrors that it should've been, that I confirmed were being passed to it.  Swapping to cache.set manually immediately fixed it.  Recall the original issue where the AWS mirror wasn't being served - this decorator was why.

PR #866, opened 2023-09-05
I want to change the mirror information I registered.

Ani-Gil:
Recently, due to the relocation of IDC, we are also trying to change the mirror server to a new
 system. Therefore, I would like to request that all information of the existing "anigil.com "
be changed to the new system "siwoo.org ".


codyro:
Both `http://mirror.siwoo.org` & `https://mirror.siwoo.org` are timing out at this time. Could you investigate?

Ani-Gil:
I think I couldn't access it because of the UTM inspection for a while. Would you like to try again?

PR #873, opened 2023-10-10
Use single redis cache key per IP

jonathanspw:
This update will move to using a single redis cache per IP, regardless of version or repository.

There will be a few major impacts from this:

- Up to 75% reduction in mirrorlist resource usage per user `dnf update` transaction.  We should almost immediately see resource usage on mirrorlist servers cut by over half upon deployment of this change.
- Up to ~240ms speedup in `dnf update` transaction for users.  (~80ms per repository after the first, accessed in the same transaction or within the same cache interval of 1 hour).  Users using our 3 default repos should see a 160ms speedup on average.
- Repos will use the same mirror in an update transaction which will alleviate an issue where each repo could hit a different mirror thus causing some update issues when an update requires packages from multiple repos and the user is served multiple mirrors which might not all be fully up to date with the latest packages depending on their sync schedules.

### DIFF
--- a/src/backend/api/handlers.py
+++ b/src/backend/api/handlers.py
@@ -24,6 +24,7 @@ from yaml_snippets.utils import (
 from api.redis import (
     _generate_redis_key_for_the_mirrors_list,
     MIRRORS_LIST_EXPIRED_TIME,
+    CACHE_EXPIRED_TIME
 )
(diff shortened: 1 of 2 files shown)


jonathanspw:
This change has had the desired effects.

image

PR #879, opened 2023-10-17
Remove mirrors.d/mirrors.qontinuum.space.yml

q0ntinuum:


codyro:
Thanks for providing a mirror! If there is anything we can help you with to reconsider, please feel free to reach out :).

PR #881, opened 2023-10-20
Use IPinfo instead of MaxMind, add logos to mirror page, serve static assets locally

### LINKED ISSUE #871: IPs without specific geolocation disproportionately hit some mirrors
Source IPs (and mirror IPs) without IP geolocation more specific than "United States, North America" which translates to coordinates "37.751, -97.822" are disproportionately hit with traffic.  This likely impacts generic geolocation points in other countries/continents as well.

We should probably add a degree of greater randomization for requests from such IPs because we truly have no idea where they are within a given country.

jonathanspw:
This update swaps out MaxMind GeoIP data for IPinfo's which resolves #871 as well as improves our geolocation data in general resulting in better mirrors being served to users.

The updates for IPinfo's data is handled externally and served via NFS because the database is quite large.

Logos for Ziplyfiber and IPinfo are added to the mirrors page.

Static assets are served locally to avoid being blocked by ad/resource blockers as well as speed up the page by minimizing external DNS requests.


ci/ansible/roles/deploy/tasks/common_setup.yml
```diff
@@ -25,6 +25,25 @@
...
+
+- name: Create ipinfo mount path
+  file:
+    path: "{{ ipinfo_path }}"
+    state: directory
+    mode: '0755'
+
+- name: Mount ipinfo path
+  ansible.posix.mount:
+    src: fs-04029e8b12637703b.efs.us-east-1.amazonaws.com:/
```
codyro, line 40: Should be a variable

jonathanspw: Fixed in forced-push.

soksanichenko (approved):

PR #919, opened 2023-10-26
Add new mirror: ftp.lip6.fr

mbouyer:
There is now a mirror at ftp.lip6.fr, per Jonathan Wright's request


jonathanspw:
Thank you!

PR #923, opened 2023-10-30
Create mirror.aarnet.edu.au.yml

tim82au:
Add mirror.aarnet.edu.au.yml (AARNet Ticket: SD-47729)

### DIFF
--- a/mirrors.d/mirror.aarnet.edu.au.yml
+++ b/mirrors.d/mirror.aarnet.edu.au.yml
@@ -1,0 +1,12 @@
+---
+name: mirror.aarnet.edu.au
+address:
+  http: http://mirror.aarnet.edu.au/pub/almalinux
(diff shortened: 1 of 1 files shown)


jonathanspw:
Thank you!

jonathanspw:
Please add the geolocation section to the mirror config.

tim82au:
Sorry about that! 

Our mirror exists in several states, is continent and country sufficient for geolocation?

jonathanspw:
> Sorry about that!
> 
> Our mirror exists in several states, is continent and country sufficient for geolocation?

So it's multiple mirrors?  If so it'd be ideal to have an entry per individual mirror, even if the URLs are identical because we route traffic using geolocation.  Assuming you're using anycast or similar having multiple in our mirror list would compliment that, not interfere with it.

tim82au:
The others are used in failover scenarios. I've added the geolocation details for the primary.

PR #927, opened 2023-10-31
Added rsync to mirrors.sonic.net

eric-eisenhart:
We added almalinux to rsyncd, so adding rsync to the address list.


jonathanspw:
Thanks!

PR #939, opened 2023-11-09
Add mirror.facebook.net mirror

davide125:
Add config for our mirror

### DIFF
--- a/mirrors.d/ash-mirror.facebook.net
+++ b/mirrors.d/ash-mirror.facebook.net
@@ -1,0 +1,15 @@
+---
+name: ash-mirror.facebook.net
+address:
+  http: http://mirror.facebook.net/almalinux/
+  https: https://mirror.facebook.net/almalinux/
+  rsync: rsync://mirror.facebook.net/almalinux/
+geolocation:
+  country: US
+  state_province: Ashburn
+  city: Virginia
+update_frequency: 3h
+sponsor: Meta Platforms Inc
+sponsor_url: https://opensource.fb.com/
+email: mirror-external@meta.com
+...
--- a/mirrors.d/snc-mirror.facebook.net
+++ b/mirrors.d/snc-mirror.facebook.net
@@ -1,0 +1,15 @@
+---
+name: snc-mirror.facebook.net
+address:
+  http: http://mirror.facebook.net/almalinux/
+  https: https://mirror.facebook.net/almalinux/
+  rsync: rsync://mirror.facebook.net/almalinux/
+geolocation:
+  country: US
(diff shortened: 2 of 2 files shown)


mirrors.d/mirror.facebook.net.yml
```diff
@@ -0,0 +1,15 @@
+---
+name: mirror.facebook.com
+address:
+  http: http://mirror.facebook.net/almalinux/
+  https: https://mirror.facebook.net/almalinux/
+  rsync: rsync://mirror.facebook.net/almalinux/
+geolocation:
+  country: US
+  state_province: Oregon
+  city: Prineville
```
davide125, line 10: This isn't entirely correct, as that endpoint is loadbalanced between two regions (Prineville, OR and Forest City, NC), but I couldn't find a way to express that here.

jonathanspw: Oh.  We have some (pretty darn good) geo-steering when we serve mirrors.  Is it possible to bypass the load balancer and just add an entry for each location?  Does the load balancer also try to geo-steer traffic?

davide125: Yeah, this is behind a load balancer that steers requests between two pools of hosts (one pool per region) and terminates SSL. We _could_ theoretically rig per-region endpoints, but that's a non-trivial endeavor and I don't know how feasible it would be in practice.

My recommendation here is to support mirrors that just declare a `country`, and setup your geosteering to send them some of the traffic directed towards the country (regardless of more specific destination).

jonathanspw: Per our chat off-list these appear to be in San Jose/Santa Clara and Ashburn with an anycasted IP based on traces.

mirrors.d/mirror.facebook.net.yml
```diff
@@ -0,0 +1,15 @@
+---
+name: mirror.facebook.com
+address:
+  http: http://mirror.facebook.net/almalinux/
+  https: https://mirror.facebook.net/almalinux/
+  rsync: rsync://mirror.facebook.net/almalinux/
+geolocation:
+  country: US
+  state_province: Oregon
```
jonathanspw, line 9:
```suggestion
  state_province: California
```

mirrors.d/mirror.facebook.net.yml
jonathanspw, line 10:
```suggestion
  city: Santa Clara
```

jonathanspw:
Thanks for setting these up!

PR #954, opened 2023-12-11
Update alma.acidman.thelefty.org.yml

metalefty:
Remove continent because it prevents 2-letter continent appear in mirrorlist.


jonathanspw:
The inconsistent continent listing is a bug, albeit minor, and will be fixed when one of us has a chance to do so.  In the meantime, there's no harm in fixing this manually just as you are doing here.

PR #975, opened 2024-01-31
Add OVHcloud mirror

sbraz:
Hi,
As discussed with @codyro and @jonathanspw, here is the pull request to add our mirror.

The IP address behind almalinux.mirrors.ovh.net uses anycast with locations in:
* Strasbourg (France)
* Roubaix (France)
* Gravelines (France)
* Beauharnois (Canada)

I set the country as "France" although it could be misleading but I don't have a good idea to make the mirror visible in Canada too.

### DIFF
--- a/mirrors.d/almalinux.mirrors.ovh.net
+++ b/mirrors.d/almalinux.mirrors.ovh.net
@@ -1,0 +1,12 @@
+---
+name: almalinux.mirrors.ovh.net
+address:
+  http: http://almalinux.mirrors.ovh.net/
+  https: https://almalinux.mirrors.ovh.net/
+geolocation:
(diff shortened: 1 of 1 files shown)


codyro:
@sbraz This is tremendous, thank you so much for providing these mirrors!

jonathanspw:
Ideally we should have 4x entries for this mirror - one for each location.  Can you open another PR with one file for each of the other 3 locations?

Big thanks for setting up the mirrors :)

sbraz:
> Ideally we should have 4x entries for this mirror - one for each location. Can you open another PR with one file for each of the other 3 locations?

Sure, see #976.

PR #976, opened 2024-02-01
OVHcloud mirror: split into 4 different locations

sbraz:
@jonathanspw does this look OK? I'm not a fan of the invalid DNS records for the names but I don't really have a better idea right now.

Our anycast IP is routed to 4 different locations. This is a follow-up to 02c86298b14.


jonathanspw:
This looks perfect.

Long term I think being able to list multiple locations per mirror would solve the potential of having to use invalid names that you did here (but is the best solution right now).

I'm not 100% sure how the system will react to multiple mirrors listing a common ASN - once an ASN match is hit I don't recall if there is any further logic on them in the case of multiple matches.  I believe we've had another recent "issue" with this.  Going to hold off on merging this until I can revisit that and make sure this won't cause any issues for your users.

jonathanspw:
All of the file names need to end in `.yml` to be picked up by the system.  It's why the existing single-mirror isn't showing up on the list.

PR #988, opened 2024-02-19
Create mirrors.hyperreal.coffee.yml

hyperreal64:
Here is my mirror information. Please let me know if you need any other information.


codyro:
Thanks for your contribution!

jonathanspw:
> Sorry, I am bad at keeping up with these things. It will not be coming back.

Thanks for the reply and letting us know and for the time you did provide a mirror.  We're happy to have you back in the future if it ever becomes feasible for you again :)

PR #993, opened 2024-03-05
Update CloudLinux private AlmaLinux repo

andrewlukoshko:
**centos.corp.cloudlinux.com** repo is not available anymore
**alma-repo.atm.svcs.io** should be used instead


mirrors.d/alma-repo.atm.svcs.io.yml
zztalker, line 1: subnets, probably should be updated too.

andrewlukoshko: Subnet is the same

PR #996, opened 2024-03-13
Adding almalinux.linux.n0c.ca to the public mirror list

infraplanethoster:
We have built this mirror to add to almalinux public mirrors. Only http is provided right now but more protocols can be offered if requested.


codyro:
Thank you for the Montreal mirror! It's up to you if you'd like to offer more protocols. For my mirrors I generally offer http/https. 

If you want to update this, make a new PR with the changes :).

PR #997, opened 2024-03-17
Update mirror.ph.net.tr.yml

mertkaracali:
Mirror has been updated, I would appreciate it if you add it to the list.


jonathanspw:
I'm not sure I understand the point of this PR.  Your site loads via https and 301s to it anyway, why serve the non-https link for the sponsor URL?

PR #998, opened 2024-03-18
Create mirror.lon.macarne.com.yml

macarnellc:
New Mirror in London


jonathanspw:
Thanks for the mirror!

macarnellc:
Hi,

Sorry guys - I messed up the firewall settings while hardening the system. Could you please re add the mirror or should we submit it again via Github? Thanks, Arne


—
Macarne LLC

codyro:
It'll get re-added to the list once the health check runs :)

PR #1002, opened 2024-03-28
Added TWDS Mirror

seadog007:
Added an new mirror


jonathanspw:
Thanks for the mirror!

PR #1005, opened 2024-04-05
add mirror almalinux.mirror.rafal.ca

rzeczkrg:
I prepared a new public mirror for AlmaLinux.


codyro:
Thank you for the mirror :)!

PR #1006, opened 2024-04-07
Add mirrors.zju.edu.cn.yml

determ1ne:


jonathanspw:
Thanks for the mirror!

PR #1007, opened 2024-04-08
Adding mirror.0x626b.com to alma mirrors list

bkolasin:
New mirror for the Alma Linux project. This PR adds the mirror to the official list.


codyro:
Thanks for the mirror, Brent :).

PR #1012, opened 2024-04-12
Add 9.4-beta

eabdullin1:


jonathanspw:
Do we need this?  We don't normally distribute betas to mirrors that I recall.

eabdullin1:
> Do we need this? We don't normally distribute betas to mirrors that I recall.

Yes. Mirrors will redirect to vault.

andrewlukoshko (approved):

PR #1018, opened 2024-04-20
Adding TheGigabit.com mirror to Almalinux

haominghm:
Adding TheGigabit.com mirror to Almalinux


codyro:
Thank you for the mirror :)!

PR #1025, opened 2024-05-01
Change name of our mirror

hcgrove:
This is a sucessor/replacement for  #1008.

I tried to comply with your request in the comments to that, even though I think I did that in the original PR. You can also see from the commit messages that I split it up (`git` insisted on treating it as a deletion and an addition as soon as I added the modified file locally), but github still shows it as a deletion and an addition. I have no idea if I can do anything to improve that, but as it is in two commits you should be able to follow it?


hcgrove:
I can see that some check (possibly the "CI/config_checker" that is mentioned) failed, but I can see what the problem is, making it quite hard to fix. When I click "Details" to see. I get a page that "details" 7 jobs, but all of them just display "Error: " when I expand them. 6 of those 7 jobs have white tickmarks, while "Run jitterbit/get-changed-files@v1" has an x on a red background (so I assume that's the one that failed, I've attached a screenshot where only that is expanded (i.e. the "Error: " for the others are hidden).

jonathanspw:
The error from CI is:

```
2024-08-22 13:15:57,608 | __main__ | ERROR | Mirror "mirror.group.one" has invalid geodata. Please check your data on "https://nominatim.openstreetmap.org/ui/details.html"
```

mirrors.d/mirror.group.one.yml
```diff
@@ -0,0 +1,16 @@
...
+  https: https://mirror.group.one/almalinux
+  rsync: rsync://mirror.group.one/almalinux
+update_frequency: 1h
+sponsor: group.one
+sponsor_url: https://www.group.one/
+email: mirror@group.one
+geolocation:
+  continent: Europe
+  country: DK
+  city: Copenhagen
```
jonathanspw, line 13:
```suggestion
  city: Copenhagen
  state_province:
```

PR #1046, opened 2024-06-07
Create mirrors.neterra.net.yaml

ktsonevski:


jonathanspw:
Thanks for the mirror!

metalefty:
@jonathanspw Hi, the yaml file is created wrong place.

ktsonevski:
Can we move it to the right place?

jonathanspw:
Fixed via https://github.com/AlmaLinux/mirrors/commit/c5a17606fe082a2a9051c6fd7ad4d077f2eb5040

Good catch @metalefty

PR #1055, opened 2024-06-18
Mirror AlmaLinux EdgeUno

kleper:
We are creating a public Mirror on AlmaLinux in Colombia and other locations in South America.

### DIFF
--- a/mirrors.d/alma.mirror.edgeuno.com.yml
+++ b/mirrors.d/alma.mirror.edgeuno.com.yml
@@ -2,7 +2,7 @@ name: mirror.one.com
 address:
   http: http://alma.mirror.edgeuno.com
   https: https://alma.mirror.edgeuno.com
-update_frequency: 24h
+update_frequency: 2h
 sponsor: EdgeUno
 sponsor_url: https://www.edgeuno.com/
 email: mirrors@edgeuno.com
--- a/mirrors.d/alma.mirror.edgeuno.com.yml
+++ b/mirrors.d/alma.mirror.edgeuno.com.yml
@@ -1,0 +1,15 @@
+name: mirror.one.com
+address:
+  http: http://alma.mirror.edgeuno.com
+  https: https://alma.mirror.edgeuno.com
+update_frequency: 24h
+sponsor: EdgeUno
+sponsor_url: https://www.edgeuno.com/
+email: mirrors@edgeuno.com
+geolocation:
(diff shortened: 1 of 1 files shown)


jonathanspw:
Thanks for setting up a mirror!

`state_province:` needs to be defined even if it is blank.  Can you please add it?

kleper:
Fixed, I add state_province.

jonathanspw:
Thanks.

I see you also defined ASN, and a quick look at your site makes me think that your ASN may represent quite a few datacenters and a geographically-diverse network.

By defining the ASN it will direct all AlmaLinux update traffic to your mirror, which in the case of geographically diverse ASNs can lead to a sub-optimal user experience.

Should it be removed or kept?

kleper:
Please kept, we are going to provide this mirror using an anycast configuration in short time, so is not going to be a problem for us.

jonathanspw:
Ok sounds great.  Sounds like you'll be adding the other locations as well?

kleper:
Yes, we will add the other locations, but our idea is this locations works over the same URL, alma.mirror.edgeuno.com

jonathanspw:
Ah one last thing, the sync frequency needs to be 3 hours or less (in the config, but in reality as well).  Syncs greater than 3 hours will results in your mirror constantly being listed as `expired` and it won't be used.

Can you update your rsync to every 3 hours or less and adjust the config here respectively?

jonathanspw:
> Yes, we will add the other locations, but our idea is this locations works over the same URL, alma.mirror.edgeuno.com

I see.  That will pose a problem as the mirror system expects a unique URL per location.  Can you create a unique URL/name for each location?  They can resolve to the same anycast IP, that part isn't a problem.

jonathanspw:
The ASN won't be able to be defined multiple times, however.  We just recently had this issue with some relevant discussions in https://github.com/AlmaLinux/mirrors/pull/1030

Since the ASN cannot be defined in multiple mirror config files the best path to force traffic in your network to your mirror(s) will be to add subnets per location.

jonathanspw:
I'd suggest dropping the ASN in the meantime until the other locations are represented, and given that ASN cannot be defined more than once.

jonathanspw:
Will you drop the ASN?  Once you do we can merge this.

jonathanspw:
@kleper friendly bump.

PR #1057, opened 2024-06-27
new mirror addition

billonenet:


jonathanspw:
Thanks for setting up a mirror!  Sync frequency needs to be every 3 hours or less.  Can you adjust this on both your rsync and in the mirror yml file?

billonenet:
Frequency adjusted to 3 hours.

jonathanspw:
Sorry I missed this at first, can you check that the sync is working properly?

http://mirrors.onenet.net/almalinux/timestamp.txt is definitely more than 3 (and 6) hours old so I think something is amiss.

PR #1064, opened 2024-07-07
Create mirror.server.net.yml

serverdotnet:
New server.net Mirror for London Location


jonathanspw:
Thanks for the mirror!

PR #1067, opened 2024-07-12
Create mirror.ox.ac.uk.yml

Jim-Leek:
University of Oxford request to host its own mirror of AlmaLinux at mirror.ox.ac.uk

### DIFF
--- a/mirrors.d/mirror.ox.ac.uk.yml
+++ b/mirrors.d/mirror.ox.ac.uk.yml
@@ -9,7 +9,7 @@ geolocation:
   country: UK
   state_province: Oxfordshire
   city: Oxford
-update_frequency: 6h
+update_frequency: 3h
 sponsor: University of Oxford
 sponsor_url: https://www.ox.ac.uk
 email: unix-team-comms@it.ox.ac.uk
--- a/mirrors.d/mirror.ox.ac.uk.yml
+++ b/mirrors.d/mirror.ox.ac.uk.yml
@@ -1,0 +1,16 @@
+---
+name: mirror.ox.ac.uk
+address:
+  http: http://mirror.ox.ac.uk/sites/repo.almalinux.org/
+  https: https://mirror.ox.ac.uk/sites/repo.almalinux.org/
+  ftp: ftp://mirror.ox.ac.uk/sites/repo.almalinux.org/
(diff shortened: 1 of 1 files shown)


mirrors.d/mirror.ox.ac.uk.yml
```diff
@@ -0,0 +1,16 @@
...
+  rsync: rsync://mirror.ox.ac.uk/sites/repo.almalinux.org/
+geolocation:
+  country: UK
+  state_province: Oxfordshire
+  city: Oxford
+update_frequency: 6h
+sponsor: University of Oxford
+sponsor_url: https://www.ox.ac.uk
+email: unix-team-comms@it.ox.ac.uk
+private: true
```
jonathanspw, line 16: With `private` set to `true` but no `asn` or `subnets` defined the mirror won't actually do anything or be used.

Do you want to force network traffic within your subnet(s) and/or ASN to your mirror first (will fall back to other mirrors if yours is down unless you set `monopoly` to `true`)?  Do you want the mirror to be public or private?

Jim-Leek: I've obviously misunderstood the meaning of 'private'.

Now hopefully set to 'false'

mirrors.d/mirror.ox.ac.uk.yml
```diff
@@ -0,0 +1,16 @@
...
+address:
+  http: http://mirror.ox.ac.uk/sites/repo.almalinux.org/
+  https: https://mirror.ox.ac.uk/sites/repo.almalinux.org/
+  ftp: ftp://mirror.ox.ac.uk/sites/repo.almalinux.org/
+  rsync: rsync://mirror.ox.ac.uk/sites/repo.almalinux.org/
+geolocation:
+  country: UK
+  state_province: Oxfordshire
+  city: Oxford
+update_frequency: 6h
```
jonathanspw, line 12: On public mirrors the sync time needs to be 3h or less.  Can you run a <=3h sync time?

Jim-Leek: OK - I think we should be ok committing to every 3h

Cheers,  Jim

jonathanspw:
Thanks so much for the mirror!

PR #1069, opened 2024-07-12
Updated Cloud Propeller official mirror, and added a backup mirror as well.

peconi:
As the title says - updating existing one (just adding Continent), and contributing a new one at a different data center (which is also behind a 200 Gbit connection to Ohio IX Internet Exchange) :-) 

Cheers.


peconi:
Oops... Fixing ASN then reopening.

PR #1070, opened 2024-07-12
Updating Cloud Propeller Mirror and adding a Backup (Secondary) Mirror as well

peconi:
As the title says - updating existing one (just adding Continent), and contributing a new one at a different data center (which is also behind a 200 Gbit connection to Ohio IX Internet Exchange) :-)

Cheers.


codyro:
Thank you! 😊

PR #1072, opened 2024-07-20
Add elmirror.cl

jkcgs:
Hello! I'm from the ElMirror.cl project team. We've been running this mirror for a while but now we want to make it public and official. If there's any change needed please say so.


jonathanspw:
Thanks for the mirror!

PR #1085, opened 2024-08-11
Create mirror.leitecastro.com.yml

tomas347:
Add mirror.leitecastro.com to the list of official mirrors


jonathanspw:
Thanks for the mirror!

PR #1088, opened 2024-08-21
Update vault mirror URL

jonathanspw:


codyro (approved):
It should probably serve on 80 as well w/o forcing a redirect (not a huge issue though).

```
bash-3.2$ curl -I vault.almalinux.org
HTTP/1.1 301 Moved Permanently
Date: Fri, 23 Aug 2024 04:07:50 GMT
Content-Type: text/html
Content-Length: 167
Connection: keep-alive
Cache-Control: max-age=3600
Expires: Fri, 23 Aug 2024 05:07:50 GMT
Location: https://vault.almalinux.org/
```

PR #1096, opened 2024-09-04
Add fr2.rpmfind.net

fbellet:


jonathanspw:
Thanks for the mirror!

PR #1098, opened 2024-09-10
Add Hoasted mirror

tommazer:
Adding mirror.hoasted.com.


jonathanspw:
Thanks for the mirror!

PR #1100, opened 2024-09-18
Create lidsol.fi-b.unam.mx.yml

ksobrenat32:
Hi!
We are a university lab focused on FOSS and are happy to include almalinux in our mirror :D


jonathanspw:
It's great to see another mirror for Mexico!  Thank you so much!

PR #1101, opened 2024-09-23
Add mirror.fiu.edu

e-rivas30:
Kindly requesting to add mirror.fiu.edu as a mirror.


e-rivas30:
> With `private` set to `true` but no `subnets` or `asns` defined the mirror will effectively do nothing (as far as our mirrorlist is conerned).
> 
> Did you mean for this to be a public mirror?

Yes, my bad. I already modified the yaml file and removed the private line.

e-rivas30 (commented):
Changes reviewed.

PR #1117, opened 2024-10-23
Add kitten to alma.acidman.thelefty.org

metalefty:


metalefty:
```
2024-10-23 05:01:35,430 | __main__ | ERROR | Mirror "alma.acidman.thelefty.org" has invalid geodata. Please check your data on "https://nominatim.openstreetmap.org/ui/details.html"
```
CI error looks false-positive.

metalefty:
I found out the cause. I'll fix the CI soon.

jonathanspw:
> I found out the cause. I'll fix the CI soon.

Thanks!

jonathanspw:
The mirror also responds for kitten on https.  Can you add that as well?

jonathanspw:
And add a `state_province` field under geo data, even if it is blank it is required.

metalefty:
In Eastern Asia, there are direct-administered municipalities by the central government, such as Taipei, that are not part of any state or province. Therefore, I was considering modifying the CI to not require a State/Province; leaving it blank should be fine.

PR #1126, opened 2024-10-29
Add Kitten to mirrors.ft.uam.es

pcolladosoto:
We just added the necessary configuration for our Kitten mirror as per [the documentation](https://wiki.almalinux.org/Mirrors.html).

If there's anything else that's needed please do let us know!

Thanks a ton for making AlmaLinux great :P


jonathanspw:
Thank you!

PR #1128, opened 2024-10-31
Add mirrors.fibianet.dk

ajodk:


jonathanspw:
the `state_province` field is required, even if blank.  The `asn` list should contain only the numbers, not the "AS" prefix.  Can you tweak these things?

ajodk:
> the `state_province` field is required, even if blank. The `asn` list should contain only the numbers, not the "AS" prefix. Can you tweak these things?

Done

jonathanspw:
Thanks for the mirror!

PR #1132, opened 2024-11-06
Create mirrors.nz.zappie.host.yml

zappiehost:
Adding a mirror in Auckland, New Zealand for AlmaLinux


jonathanspw:
Thanks for the mirror!

PR #1133, opened 2024-11-09
Added HyperBit SRLs Italian Mirror

francescomasala:
Added new mirror:
- 40 Gbps bandwidth
- IPv4 
- IPv6

Ps: The mirror is finishing the sync stage


jonathanspw:
Thanks for the mirror!

Issue #1143, opened 2024-11-18
Mirrorlist should deal with unmatched, slightly incorrect versions and multiple minors for one major


jonathanspw:
This came up while releasing 9.5 - the mirror system currently cannot handle 9.5 and 9.4 being support versions at the same time because only one exists in the `duplicated_versions` mapping.  This is a regression introduced by the kitten support PR.

Relevant points in code and errors:

https://github.com/AlmaLinux/mirrors/blob/mirrors_service/src/backend/api/handlers.py#L434
https://github.com/AlmaLinux/mirrors/blob/mirrors_service/src/backend/api/handlers.py#L434

```
/opt/mirror_service/src/backend/update_mirrors.py:272: DeprecationWarning: There is no current event loop
  loop = asyncio.get_event_loop()
Traceback (most recent call last):
  File "/opt/mirror_service/src/backend/update_mirrors.py", line 273, in <module>
    loop.run_until_complete(update_mirrors())
  File "/usr/lib64/python3.12/asyncio/base_events.py", line 684, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/opt/mirror_service/src/backend/update_mirrors.py", line 269, in update_mirrors
    await task
  File "/opt/mirror_service/src/backend/update_mirrors.py", line 103, in update_mirrors_handler
    mirror_iso_uris = mirror_processor.get_mirror_iso_uris(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/mirror_service/src/backend/api/mirror_processor.py", line 472, in get_mirror_iso_uris
    for arch in arches[base_version]:
                ~~~~~~^^^^^^^^^^^^^^
KeyError: '9.5'
```

The expected functionality is for versions without a mapping in duplicate_version to still attempt to map to a valid major.

For example:

9.2 should map to 9
9.1111 should map to 9
9.97234 should map to 9

These are all past-supported versions, or invalid minors, but we should map the pattern to the major where we can make reasonable assumptions.

PR #1154, opened 2024-12-05
Solves AlmaLinux/mirrors#1143

soksanichenko:
- Check if SELinux is enabled during Ansible deploying
- PEP8 (long lines, indents, typing etc.)
- The unused import are removed, the rest imports are optimized
- Get right major version if a full version is not listed in duplicated versions (the main issue)
- Old field `versions_arches` is removed from the service config. Field `arches` provides the full compatibility with it

### DIFF
--- a/src/backend/main.py
+++ b/src/backend/main.py
@@ -17,11 +17,12 @@ from api.handlers import (
     get_all_mirrors,
     get_isos_list_by_countries,
     get_main_isos_table,
-    get_allowed_arch,
     check_optional_version,
     get_optional_module_from_version,
     SERVICE_CONFIG_JSON_SCHEMA_DIR_PATH,
-    SERVICE_CONFIG_PATH, get_allowed_version, get_allowed_arch,
+    SERVICE_CONFIG_PATH,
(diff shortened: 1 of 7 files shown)


src/backend/main.py
```diff
@@ -164,22 +166,38 @@ def get_mirror_list(
     
     # protocol get arg
     request_protocol = request.args.get('protocol')
-    if request_protocol and request_protocol not in ["http","https"]:
-        return "Invalid input for protocol, valid options: http, https"
+    if request_protocol and request_protocol not in ['http', 'https']:
```
gost-serb, line 169: Wouldn't 
```suggestion
    if request_protocol not in ['http', 'https']:
```
do the same?

gost-serb (approved):

src/backend/main.py
soksanichenko: No. Because we return any mirrors if request_protocol is None

jonathanspw (approved):
LGTM, thanks for the cleanup in many areas.

PR #1155, opened 2024-12-05
Solves AlmaLinux/mirrors#1143

soksanichenko:
- Check if SELinux is enabled during Ansible deploying
- PEP8 (long lines, indents, typing etc.)
- The unused import are removed, the rest imports are optimized
- Get right major version if a full version is not listed in duplicated versions (the main issue)
- Old field `versions_arches` is removed from the service config. Field `arches` provides the full compatibility with it

### DIFF
--- a/data_models.py
+++ b/data_models.py
@@ -94,7 +94,7 @@ class MirrorData:
     asn: list[str] = None
     monopoly: bool = False
     urls: dict[str, str] = field(default_factory=dict)
-    module_urls: dict[str, list] = field(default_factory=dict)
+    module_urls: dict[str, dict] = field(default_factory=dict)
     has_optional_modules: Optional[str] = None
     subnets: list[str] = field(default_factory=list)
     subnets_int: list[tuple] = field(default_factory=tuple)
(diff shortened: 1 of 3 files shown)


utils.py
```diff
@@ -505,11 +523,11 @@ def _get_arches_for_version(
...
-        versions_arches: dict[str, list[str]]
+        arches: dict[str, list[str]]
 ) -> bool:
-    if version not in versions_arches:
+    if version not in arches:
         return True
-    elif version in versions_arches and arch in versions_arches[version]:
+    elif version in arches and arch in arches[version]:
         return True
     else:
```
gost-serb, line 532: No need for `else`. `return False` will do the same

gost-serb (approved):

data_models.py
```diff
@@ -152,12 +152,11 @@ class MainConfig:
...
     versions: list[str] = field(default_factory=list)
-    optional_module_versions: dict[list] = field(default_factory=dict)
-    arches: list = field(default_factory=list)
+    optional_module_versions: dict[str, list[str]] = field(
+        default_factory=dict
+    )
+    arches: dict[str, list] = field(default_factory=dict)
     duplicated_versions: dict[str, str] = field(default_factory=dict)
     vault_versions: list[str] = field(default_factory=list)
-    versions_arches: dict[str, list[str]] = field(
```
jonathanspw, line 159: Per our chat, this is fine for the 10-beta but it removes the ability to pin repo support between minor versions and architectures for optional modules.  This will likely be needed for EPEL10 on v2 if we choose to rebuild it for v2 so we'll have to revisit this in the near future.

jonathanspw (approved):
This one LGTM, about to work on the associated PR in https://github.com/AlmaLinux/mirrors/pull/1154/files

PR #1156, opened 2024-12-06
Add new mirror for Shiv Nadar Institution of Eminence

bhuvnesh-snioe:
Add new mirror for Shiv Nadar Institution of Eminence

### DIFF
--- a/mirrors.d/mirror.snu.edu.in.yml
+++ b/mirrors.d/mirror.snu.edu.in.yml
@@ -1,0 +1,14 @@
+---
+name: mirror.snu.du.in
+address:
+  http: http://mirror.snu.edu.in/almalinux/
+  https: https://mirror.snu.edu.in/almalinux/
+geolocation:
+  country: IN
+  state_province: Uttar Pradesh
+  city: Dankaur
+update_frequency: 3h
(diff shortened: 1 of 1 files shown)


jonathanspw:
Thanks for the mirror!  I see that the DNS resolves to Cloudflare which isn't really ideal.  Can you point it directly at the mirror?

bhuvnesh-snioe:
> Thanks for the mirror! I see that the DNS resolves to Cloudflare which isn't really ideal. Can you point it directly at the mirror?

Hi,

We have changed the necessary changes. Now it's resolving the mirror server IP.

jonathanspw:
Looks great!  I notice that your organization seems to have its own ASN: AS132785

Would you like to configure our mirrorlist to send AlmaLinux mirrorlist requests from within your ASN to your mirror?

This is easily done by adding `asn: 132785` to the yaml config.

bhuvnesh-snioe:
> Looks great! I notice that your organization seems to have its own ASN: AS132785
> 
> Would you like to configure our mirrorlist to send AlmaLinux mirrorlist requests from within your ASN to your mirror?
> 
> This is easily done by adding `asn: 132785` to the yaml config.

Hi,

We have updated the asn number

PR #1157, opened 2024-12-09
Create mirror.telepoint.bg.yml

Cremator:


jonathanspw:
Thanks for the mirror!  Would you like traffic within your ASN to automatically be direct to your mirror?

This can be done by adding `asn: 31083` to the config.  This is totally optional based on your desire.

If the ASN represents more than one physical network location this might be sub-optimal.

Cremator:
Updated the yaml with ASN info.

PR #1167, opened 2024-12-25
Add xicalcoatl.mx mirror

maidlinux01:
Hello 
I just create my own mirror, and I want to contribute to this awesome Linux community

### DIFF
--- a/mirrors.d/linux-mirror.xicalcoatl.mx.yml
+++ b/mirrors.d/linux-mirror.xicalcoatl.mx.yml
@@ -1,0 +1,17 @@
+---
+name: linux-mirror.xicalcoatl.mx
+address:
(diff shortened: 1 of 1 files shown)


jonathanspw:
Hi!

Thanks so much for the mirror.  Mexico is definitely an underserved area mirror-wise.

This appears to be on a residential connection with a dynamic IP (which is fine).  What sort of speeds can this connection support?

maidlinux01:
Hi!
I'm glad to help to grow this project.
This one is a fiber connection; right now it can support up to 125 Mbps speeds, which could increase in the future.

jonathanspw:
If it increases in the future all the better :D

Thanks again for the mirror!

PR #1180, opened 2025-02-14
Create mirror.ixirhost.com.yml

hlldmr:


jonathanspw:
Thanks for the mirror!

Can you please use a sync frequency of 3 hours or less?

hlldmr:
Mirror is active, but not listed yet, is something missing?

PR #1191, opened 2025-03-17
Add mirror.a1.mk

a1filiptodosov:
Added new mirror:

10 Gbps bandwidth
IPv4
IPv6


a1filiptodosov:
@jonathanspw please let me know if there are any problems with the configuration yml. Thanks.

jonathanspw:
Thanks for the mirror submission!  Can you adjust the sync frequency to be every 3 hours or less?

PR #1195, opened 2025-03-21
Update atl.mirrors.ezscale.cloud.yml

Prophet731:
Added FTP availability


Prophet731:
Added AlmaLinux Kitten to mirror

PR #1201, opened 2025-04-26
Update mirrors.isu.net.sa.yml to be mirror.maeen.sa.yml

sysdev-isu:
Dear Team,

We migrated our servers to new location with higher specs and added kitten.

please update the old mirror with the new details to disable it.

BR


sysdev-isu:
Hi Team,

Please check this and update the mirror list the geo data is correct but the config check didnt pass

BR

PR #1207, opened 2025-05-08
Update mirror.nodespace.com

tnewton1:
The NodeSpace mirror is back online. Switching the domain from the legacy nodespace.net to current nodespace.com.


jonathanspw:
Hi,

Thanks for bringing the mirror back online!  It appears outdated right now.  Can you please verify that the rsync cron/task is working as expected?

tnewton1:
Yep - sorry about that. I was on autopilot, copied the command right out of the wiki and needed to adjust it. Should be correct now. Thanks!

PR #1223, opened 2025-05-27
Create vn-mirrors.techhost.vn.yml

techhostvn:


jonathanspw:
Hi, thanks for the mirror!

Please add the geolocation fields.

jonathanspw:
Thanks for that.  One more thing, your rsync looks incomplete.  There are files missing from the root /almalinux directory (for example `TIME` and `timestamp.txt`, some GPG keys, among others.

Are you doing some sort of filters or excludes in your rsync?

techhostvn:
Hi, i was update it

PR #1235, opened 2025-07-07
Create mirrors.tino.vn.yml

LeAnhlinux:


jonathanspw:
Is this different from https://github.com/AlmaLinux/mirrors/blob/master/mirrors.d/mirrors.tino.org.yml ?

jonathanspw:
@LeAnhlinux

PR #1241, opened 2025-08-04
Create mirror.africloud.com2.yml

niyyie:
Update to a previous listing, correcting minor mistakes with mirror name and sponsor name.


jonathanspw:
You need to modify the other one, not make a duplicate.

PR #1256, opened 2025-09-04
Create almalinux.elhacker.net.yml

ehnwebmaster:


jonathanspw:
Howdy and thanks for the mirror.  The mirror dataset appears to be incomplete.  Can you please verify if the sync is working properly?  Are you using any excludes?

ehnwebmaster:
The sync is completed

> du -hs /homes/almalinux
1.5T	/homes/almalinux

PR #1258, opened 2025-09-06
Create alma.c-servers.co.uk.yml

alfatarsos:
Kindly note that, on FTP, also FTPS will be provided.

### DIFF
--- a/mirrors.d/mirror.almalinux.peralex.com.yml
+++ b/mirrors.d/mirror.almalinux.peralex.com.yml
@@ -1,0 +1,17 @@
+---
+name: mirror.almalinux.peralex.com
+address:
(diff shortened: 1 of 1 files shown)


jonathanspw:
Howdy, thanks for the mirror.  It looks like your mirror is outdated by a few days.  Can you ensure that the rsync is setup properly?

I also notice that the location that this DNS/IP actually resolve to appears to be Atlanta, GA, not Montreal - perhaps Bunny CDN?  That could explain why I see Atlanta.  Can you confirm where the server is actually hosted?

Mirrors should generally use a direct address to the server and not serve content through a CDN.  Our mirror system is designed to handle the traffic routing aspect of things and adding in CDN layers doesn't play nice with that.

PR #1266, opened 2025-09-09
Create mirror.vpspal.com.yaml

vps-pal:
Hello
Please add mirror
Best

### DIFF
--- a/mirrors.d/mirror.vpspal.com.yaml
+++ b/mirrors.d/mirror.vpspal.com.yaml
@@ -1,0 +1,19 @@
+---
+name: mirror.vpspal.com
+address:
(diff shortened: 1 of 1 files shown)


jonathanspw:
Is this different from https://github.com/AlmaLinux/mirrors/pull/1264 ?

vps-pal:
Yes, it's different
I just copied it

jonathanspw:
I mean behind the scenes is anything actually different?  The mirrors have different IPs but go to the exact same network and the websites look pretty much identical too.

vps-pal:
Both are on different servers but the network is the same 
Our network has enough capacity to host both servers 
Is this a problem?

PR #1268, opened 2025-09-12
Create almalinux.mirror.archettitechnology.it.yml

archettitechnology:
i want to join on almalinux mirrors.

Cheers


jonathanspw:
Thanks for the mirror!

PR #1274, opened 2025-09-18
Create alma

doctatortot:
new prospective mirror for almalinux/kitten - Montreal QC Canada
stygian.failzero.net/mirror/alma
stygian.failzero.net/mirror/almakitten


mirrors.d/stygian.failzero.net.yml
```diff
@@ -0,0 +1,21 @@
+name: stygian.failzero.net/mirror/alma
```
jonathanspw, line 1:
```suggestion
name: stygian.failzero.net
```

PR #1276, opened 2025-09-20
Bring back CICKU mirror

cicku:
Bring back the US location only at the moment.


jonathanspw:
Are you sure it's online?

```
$ curl https://mirror.cicku.me
curl: (6) Could not resolve host: mirror.cicku.me
```

cicku:
Fixed!

PR #1278, opened 2025-09-29
New mirror in South Korea!

wonchandev:
Hello

Jeonnam High School in South Korea supports Almalinux mirroring services.


jonathanspw:
Thanks for the mirror!

PR #1336, opened 2026-03-07
Create mirror.techrich.hk.yml

rttnewuseguy:
adding our mirror

### DIFF
--- a/mirrors.d/mirror.techrich.hk.yml
+++ b/mirrors.d/mirror.techrich.hk.yml
@@ -1,0 +1,15 @@
+---
+name: mirror.techrich.hk
+address:
+  http: http://mirror.techrich.hk/almalinux/
+  https: https://mirror.techrich.hk/almalinux/
+geolocation:
+  country: US
+  state_province: Illinois
+  city: Chicago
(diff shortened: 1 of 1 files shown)


jonathanspw:
Howdy and thanks for the mirror.  It appears to have 3 A records behind it pointing to 3 different geographical areas.  Is this some sort of CDN?

We only accept direct single-location mirrors at this time as we do the geographic coordination within our system and adding CDNs to this has a negative impact on the resulting mirrors served to users.

jonathanspw:
@rttnewuseguy any updates on this?

jonathanspw:
@rttnewuseguy

rttnewuseguy:
Hi Jonathan, sorry this account isn't monitored often and was created just to announce the mirror.  We have multiple servers to increase throughput.  One thing we can consider in the future is combining the servers to a single area or we may be able to make a separate mirror hostname for Almalinux that just resolves to a single server.

jonathanspw:
Makes sense.  We definitely need one physical mirror per mirror entry with few exceptions for large networks.  Our mirror system does the balancing accordingly.

PR #1337, opened 2026-03-09
Add mirror.rabisu.com (DE)

mirrorrabisu:
New mirror submission - Mirror Rabisu - mirror.rabisu.com - Germany

### DIFF
--- a/mirrors.d/de/mirror.rabisu.com.yaml
+++ b/mirrors.d/de/mirror.rabisu.com.yaml
@@ -1,0 +1,18 @@
+---
+name: Rabisu
+website: https://www.mirror.rabisu.com/
+sponsor:
+  name: Rabisu
+  url: https://www.rabisu.com/
+location:
+  country: DE
+  city: Falkenstein
+urls:
+  http: http://mirror.rabisu.com/mirrors/almalinux/
+  https: https://mirror.rabisu.com/mirrors/almalinux/
+  rsync: rsync://mirror.rabisu.com/mirrors/almalinux/
+protocols:
+  - http
+  - https
+  - rsync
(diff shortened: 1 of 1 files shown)


jonathanspw:
Hi and thanks for the mirror!  The yaml config you submitted doesn't follow the proper structure so the mirror wouldn't be included in our list.

Please use https://github.com/AlmaLinux/mirrors/blob/master/mirrors.d/repo.almalinux.org.yml as a formatting example and/or reference https://wiki.almalinux.org/Mirrors.html for more complete documentation on the yaml config file.

mirrorrabisu:
Could you please double-check? I've updated it correctly.

mirrors.d/de/mirror.rabisu.com.yaml
```diff
@@ -0,0 +1,15 @@
...
+name: mirror.rabisu.com
+address:
+  http:  http://mirror.rabisu.com/mirrors/almalinux/
+  https: https://mirror.rabisu.com/mirrors/almalinux/
+  rsync: rsync://mirror.rabisu.com/mirrors/almalinux/
+geolocation:
+  country: DE
+  state_province: SN
+  city: Falkenstein
+update_frequency: Twice A Day
```
jonathanspw, line 11:
```suggestion
update_frequency: 12h
```

jonathanspw (changes requested):
`12h` is the proper way to represent this.

Ideally sync frequency is every `3h`. Would it be possible for you to sync more often than every 12h?

jonathanspw:
@mirrorrabisu

jonathanspw:
@mirrorrabisu will you be able to implement the suggestions?

PR #1342, opened 2026-04-15
Remove mirror.plusserver.com

tpeltzer:
Unfortunately, we will be removing our public AlmaLinux mirror


jonathanspw:
Thanks for the time you ran it!  We'd love to have it back in the future if you can host it again in the future.

PR #1359, opened 2026-04-25
Add mirror.plusline.net

jnk0:
This pull request adds the mirror server “mirror.plusline.net”, which is based in Frankfurt am Main, Germany.


jonathanspw:
Thanks for the mirror!

jonathanspw:
It looks like your mirror is out of date and missing some files from the main directory.  Can you double check that your sync is working properly?

jnk0:
Double-checked and flipped the bits. Everything should be fine now.

PR #1364, opened 2026-05-04
Fix missing trailing slashes in mirror.tzulo.com URLs

cereal7802:
Add trailing slashes to https and rsync addresses.

### DIFF
--- a/mirrors.d/mirror.tzulo.com.yml
+++ b/mirrors.d/mirror.tzulo.com.yml
@@ -2,8 +2,8 @@
 name: mirror.tzulo.com  
 address:
   http: http://mirror.tzulo.com/almalinux/
-  https: https://mirror.tzulo.com/almalinux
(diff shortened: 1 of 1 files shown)


jonathanspw:
Howdy.  This config doesn't require or need the trailing slashes - we append those as needed in the codebase.

I see that your mirror is marked as invalid, but it's because of missing content.  For example this URL should be valid but it 404s: http://mirror.tzulo.com/almalinux/8.10/AppStream/aarch64/os/repodata/repomd.xml

Can you check on your sync process/cron?

cereal7802:
Hello,


This was corrected. You should see all arch included now. The wrong sync 
script was being used and it was limiting the mirror files to x86_64 only.

jonathanspw:
Your mirror is in the list now :)

I'll close this.  Thanks for fixing the sync.

PR #1368, opened 2026-05-14
Add www.maple-hosting.com mirror (Netherlands)

morhsn:


jonathanspw:
Thanks for the mirror!  It looks like you're using Cloudflare for your domain.  We don't allow CDNs to be part of our mirror system as the mirror system essentially is a CDN.

morhsn:
Thanks for the quick review and for clarifying.
The mirror content is served from a dedicated 10Gbps bare metal server in
Lelystad, Netherlands with ~7TB storage — Cloudflare is not caching any of
the mirror content. Cache is explicitly bypassed for all /mirror/* paths,
and the Cloudflare Worker simply proxies requests straight through to the
origin server in real time with no CDN behavior involved.
Would you be able to confirm whether this setup is acceptable given that
there's no actual CDN caching taking place, and traffic is passing directly
to the dedicated mirror server?

Mor Hasson

jonathanspw:
Would you be able to put the mirror on a subdomain?  That'd be ideal.

The system has fallbacks for geolocation on a mirror's IPs if certain other things fall over and Cloudflare IPs make a mess.

Additionally, even without caching traffic Cloudflare is still in the data path in this case.

morhsn:
Done, please check. Thank you!

jonathanspw:
Thanks for the mirror!

PR #1377, opened 2026-06-05
removal of the mirror mirrors.atlas.net.co from the list

marcomuskus:


jonathanspw:
Thanks for the time you ran the mirror.  We'd love to have you back if you're able to run it again in the future.

